PR #93, opened 2016-09-27
Add DocumentStore

### LINKED ISSUE #56: DocumentStore
Add a new datastore to orbit-db: Document Store. 

Original discussion here: https://github.com/haadcode/orbit-db/issues/35

shamb0t:
DocumentStore to retrieve docs by string contained in title
closes #56

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,8 @@
     "orbit-db-eventstore": "0.1.5",
     "orbit-db-feedstore": "0.1.4",
     "orbit-db-kvstore": "0.1.3",
-    "orbit-db-pubsub": "0.0.4"
+    "orbit-db-pubsub": "0.0.4",
+    "orbit-db-docstore": "0.0.7"
   },
   "devDependencies": {
     "asyncawait": "^1.0.6",
--- a/src/OrbitDB.js
+++ b/src/OrbitDB.js
@@ -5,6 +5,7 @@ const EventStore    = require('orbit-db-eventstore')
 const FeedStore     = require('orbit-db-feedstore')
 const KeyValueStore = require('orbit-db-kvstore')
 const CounterStore  = require('orbit-db-counterstore')
+const DocumentStore = require('orbit-db-docstore')
 const Pubsub        = require('orbit-db-pubsub')
 const Cache         = require('./Cache')
 
@@ -35,6 +36,10 @@ class OrbitDB {
 
   counter(dbname, options) {
     return this._createStore(CounterStore, dbname, options)
+  }
+
+  docstore(dbname, options) {
+    return this._createStore(DocumentStore, dbname, options)
   }
 
   disconnect() {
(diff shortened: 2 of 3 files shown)


haadcode:
Sorry @shamb0t I haven't had time yet to look at this. Will do so this week, promise! :)

shamb0t:
@haadcode tyt! :)

haadcode:
OK, took a look. Generally it's exactly what we want, good job! :+1:

Would you want to add the possibility to define the key you index by? Something like:
```javascript
constructor(ipfs, id, dbname, options) {
...
this.options.indexByKey = options.indexBy // options = { indexBy: '_id' }, ie. use the _id field of the doc in put(doc) as the key to index by.
...
}

  put(doc) {
    return this._addOperation({
      op: 'PUT',
      key: doc[this.options.indexBy],
      value: doc,
      meta: {
        ts: new Date().getTime()
      }
    });
  }
```

haadcode:
Another cool option we should add is the possibility to use a mapper (a map function) to query the documents. See https://pouchdb.com/guides/queries.html, they have very good documentation. The idea is that the consumer can provide a function through which the documents are filtered when queried.

So in our case, we could add a new method to the Store called `query(mapper)` that would be then called with `store.query((doc) => doc.someField === 17)`. This passes the function to the store and inside `query(doc)` method, we can do:
```javascript
query(mapper) {
    return Object.keys(this._index._index)
      .map((e) => this._index.get(e)) // get all documents in the index
      .filter((e) => mapper(e)) // e is the document, run it through mapper (which returns `doc.someField === 17` true|false
}
```

Does that make sense?

This is not an optimized solution but would get us started to test it and figure out the exact API. If you have any ideas for the naming/API convention, feel free to use better names for the functions and params.

shamb0t:
@haadcode updated the pr!

It currently indexes by '_id' if no index key is specified

haadcode:
The package that this PR will add is here: https://github.com/shamb0t/orbit-db-docstore

(for my own convenience, always forget that it's a separate repo :))

haadcode:
@shamb0t Looks good! Could you provide an example usage of the query/mapper in the README? 

I think otherwise this is good to merge code-wise.

I'll need to test this before I can merge it in and I don't have the bandwidth this week unfortunately (greetings from Toronto!). What would make it easier and quicker, would be to write tests but they're definitely NOT required. It's just that I want to make sure that this works as expected after it's merged and right now I have to do it manually. If you do want to write tests, and by all means you can say no, you could do that directly in orbit-db (that's where I test the other stores), but my plan is to move the tests of the individual stores to their respective repos eventually.

Great work so far! Looking forward to have a document store in orbit-db :) Thanks @shamb0t.

shamb0t:
@haadcode added some tests!

haadcode:
Excellent! Can you take a look at https://circleci.com/gh/haadcode/orbit-db/495?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link and see why it errors?

shamb0t:
Yea I messed up the rebase with the mocha update...should be good now!

haadcode:
Fantastic! Great work @shamb0t, this all looks beautiful :)

Going to merge right now. Thank you so much for implementing this!

PR #112, opened 2016-10-20
feat: use localstorage for caching

dignifiedquire:
@haadcode  :)


src/Cache.js
```diff
@@ -1,55 +1,38 @@
...
-        resolve()
-      }
-    })
-  }
+let store = typeof window !== 'undefined' && window.localStorage
+
+if (!store) {
+  const LocalStorage = require('node-localstorage').LocalStorage
+  store = new LocalStorage('./scratch')
+}
```
haadcode: We should be able to define the path where the data is saved. 

Perhaps a solution here would be to create the instance in loadCache() and use filePath - (filename.ext) as the data directory, and filename.ext as the key?

haadcode: Or pass the data path as a param in loadCache()

dignifiedquire: sure, do you want to be it different to the `cacheFile` or should I just reuse that?

haadcode: Re-use `cacheFile` param I guess. loadCache(cacheFile) should be the full path of the file. from there, you can separate the file and path from `cacheFile`. loadCache should be only called once before cache is ready to use (ie. don't have to pass the cache file path to get/set) . Does that make sense?

dignifiedquire:
@haadcode updated, but the tests are failing now. It seems the reason is that there is always a cachefile defined and something in the tests doesn't get cleaned up properly. Could you take a look please?

haadcode:
As discussed, let's not use local storage but stick to the previous solution or use blob-stores. Closing this one @dignifiedquire.

PR #146, opened 2016-11-26
Fixup makefile

Kubuxu:
This makes it safe it `make` is run with -j X.

It was derp on my side as I had MAKEFLAGS set.


haadcode:
Thanks @Kubuxu! 👍

PR #225, opened 2017-07-17
Fix docstore function links

atfornes:
Fix empty links in docstore functions


haadcode:
This is great, thank you for fixing (writing) the links!

PR #247, opened 2017-11-20
Use js-ipfs@0.27.0

haadcode:


haadcode:
Moved to https://github.com/orbitdb/orbit-db/pull/247

PR #262, opened 2017-12-02
Fix replication starting from wrong heads

### LINKED ISSUE #258: Can't replicate different, multiple databases (event concurrency bug?)
This is an issue I have while trying to replicate two different dbs (in my case a keyvalue and a feed):

```javascript
async function replicateFeed(address) {
    const feedDb = await orbitdb.feed(address)
    feedDb.events.on('replicated', () => console.log('feed replicated)', address)
    feedDb.events.on('replicate', () => console.log('feed replicate)', address)
    feedDb.events.on('replicate.progress', console.log('feed progress)', address)
}

async function replicateKvalue(address) {
    const feedDb = await orbitdb.keyvalue(address)
    feedDb.events.on('replicated', () => console.log('keyvalue replicated)', address)
    feedDb.events.on('replicate', () => console.log('keyvalue replicate)', address)
    feedDb.events.on('replicate.progress', console.log('keyvalue progress)', address)
}

await replicateKvalue(addressOfKvalue)
await replicateFeed(addressOfFeed)
```

If I run the above code, ~I get only the feed events. If I invert the bottom two lines, I get only the keyvalue events~ I get only the feed callback. The events, however, are sometimes the feed events and sometimes the keyvalue ones (ie, 50% of the time I get all the feeds and 50% all the keyvalues, despite the callbacks being called being only the feed one). If I invert the lines, vice versa.

Note that `replicateKvalue` and `replicateFeed` are **not** running concurrently: I `await` for each before running the other. What runs concurrently is the replication in the background. For some reason, only the last replication events seem to "win", but the updates are random. Looks like event rewriting?

In my case, waiting for replication serially (ie calling `replicateFeed` in the feed's `replicated` event handler) is not an option: I'm trying to write the "Twitter" demo and I'm doing these two replications for each followed user: If user are many, waiting serially for each of their databases and then waiting serially for each one of the users will take a lot of time.

haadcode:
This PR will fix the bug where replication starts from wrong heads when multiple databases are in use at once.

Use ipfs-pubsub-room branch with a fix for the wrong heads
Add a test tocheck that the automatic replication receives the right heads
Add more logging

Closes #258


cristiano-belloni:
Thanks, @haadcode, will try that asap (I'm out all day)

PR #279, opened 2017-12-11
Replication fixes

haadcode:
This PR will:
- Fix replication issues
- Add tests to make sure a database can be saved and loaded multiple times in a row
- Add a test to make sure the correct replication progress is emitted
- Add mem-store for faster development testing
- Improve tests
- Improve logging
- Fix replication benchmark
- Update packages and build


haadcode:
Closes #285

PR #282, opened 2017-12-13
Fix browser build

### LINKED ISSUE #280: TypeError: exists is not a function
Hi,
It seems that in 'orbit-db-cache' you require 'level', which in turn requires 'node-bindings'. 
'node-bindings' has a bug and returns the above error in a clean orbit-db install.

Luckily it seems they fixed it: https://github.com/TooTallNate/node-bindings/pull/20 + https://github.com/Level/level/issues/34

It might be just a matter of updating the version of `level`? Right now, installing `orbit-db` and importing it dies.

### LINKED ISSUE #285: Unable to replicate a db from its address (Chrome, orbit-db@0.18.10, ipfs@0.27.4)
Hi,

In one instance of my app I have a user creating a feed and sharing the feed's address.
In another instance (another browser), I have a user importing the address manually and trying to replicate the feed. The issue I'm experiencing at the moment is that the feed is never replicated and waits indefinitely:

    // This never returns
    const feedDb = await orbitdb.feed(address)

Please note that I'm trying this with `orbit-db@0.18.10`, because of https://github.com/orbitdb/orbit-db/issues/280, on Chrome. If it's a known bug that should disappear with a newest version, I'll just wait for #280 to be closed.

Also note that I *was* able to replicate databases with the same code. The last version I had working was `orbit-db@0.18.3`.

haadcode:
This PR will:
- Add a build for browsers to include with `import OrbitDB from 'orbit-db'`
- Add a webpack config file to build ^ without `leveldown` (LevelDB)
- Set package.json's `browser` field to `dist/orbit.js` built with ^
- Clean up webpack config files and remove all obsolete settings
- Pull in a fix from orbit-db-cache

Closes #280
Closes #285


haadcode:
Will merge this soon, update the changelog and release as 0.19.0 before midnight today :)

cristiano-belloni:
Thanks! Still travelling, but will try as soon as I come back home :)

PR #286, opened 2017-12-23
Fix/doc and examples update

ricardojmendez:
A couple of small fixes to the build process and documentation that I found while testing it out. See commits for details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,13 @@ cd orbit-db
 npm install
 ```
 
+You'll also need babel and webpack, if you don't have them installed already:
+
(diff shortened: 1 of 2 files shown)


haadcode:
Thanks @ricardojmendez for this PR! These kind of things are, while small, hugely important for the developer experience, so your work here is highly appreciated! ❤️ 

Welcome to OrbitDB contributors 😃

ricardojmendez:
Glad I can help and thanks for the welcome! I'll let you know if I run into anything else.

PR #304, opened 2018-01-03
Change "the only" to "only the".

lufte:


haadcode:
Thank @lufte, good catch! 👍 😄

Issue #307, opened 2018-01-08
Custom database types


garbados:
Although the orbit-db-store documentation suggests you can use custom stores (["Creating Custom Data Stores"](https://github.com/orbitdb/orbit-db-store#creating-custom-data-stores)), doing so is actually quite difficult because of how OrbitDB [pre-processes](https://github.com/orbitdb/orbit-db/blob/master/src/OrbitDB.js#L73) new stores. Indeed, OrbitDB [rejects](https://github.com/orbitdb/orbit-db/blob/master/src/OrbitDB.js#L104) stores with types not matching those it doesn't already know about.

In [cubesat](https://github.com/garbados/cubesat), I've taken to [piggy-backing off of an existing store type](https://github.com/garbados/cubesat/blob/master/index.js#L109-L115) in order to use a custom store with a custom index. My approach is hacky as heck and I'd love a better way.

haadcode:
Indeed, you're right! Adding stores currently is cumbersome. Your PR is a great fix for it, so let's do that :)

Btw. if you're interested, we have https://github.com/orbitdb/orbit-db/issues/4 open and looking for contributors. Take a look and see if you'd be interested. I think it'd be the best way to solve what you're doing in ^ and would love to have a "PouchDBStore" in OrbitDB :)

garbados:
Yay! I've jumped into that thread and I hope to have something together for it shortly.

haadcode:
This should now be fixed through the PR above (#308). @garbados if you feel this can be closed, feel free to do so or if there's still something that needs work, let's continue the discussion.

Thank you for all the contributions you've made around this!

garbados:
Yay! Seems good to me 👍

PR #308, opened 2018-01-08
Add custom types

garbados:
Add an `.addType(name, store)` method that updates the `databaseTypes` object so that OrbitDB can pre-process them like other store types.

Addresses https://github.com/orbitdb/orbit-db/issues/307

### DIFF
--- a/src/OrbitDB.js
+++ b/src/OrbitDB.js
@@ -18,7 +18,7 @@ const logger = Logger.create("orbit-db")
 Logger.setLogLevel('NONE')
 
 // Mapping for 'database type' -> Class
-const databaseTypes = {
+let databaseTypes = {
   'counter': CounterStore,
   'eventlog': EventStore,
   'feed': FeedStore,
@@ -343,6 +343,11 @@ class OrbitDB {
 
   static isValidType (type) {
     return Object.keys(databaseTypes).includes(type)
+  }
+
+  static addDatabaseType (type, store) {
+    if (databaseTypes[type]) throw new Error(`Type already exists: ${type}`)
+    databaseTypes[type] = store
   }
 
   static create () {
--- a/test/create-type.test.js
+++ b/test/create-type.test.js
@@ -1,0 +1,59 @@
+'use strict'
+
+const assert = require('assert')
+const config = require('./utils/config')
+const DocumentStore = require('orbit-db-docstore')
+const OrbitDB = require('../src/OrbitDB')
+const rmrf = require('rimraf')
+const startIpfs = require('./utils/start-ipfs')
+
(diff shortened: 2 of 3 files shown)


haadcode:
Hi @garbados! Thank you for the PR, looks great and would be happy to add the functionality it brings. You're absolutely right in #307 that it's cumbersome atm.

There's couple a small comments for the PR that we should address before merging. Also, I appreciate the very clean commit style, but could you squash the commit to one before we merge (you can put all three commit messages into the squashed one)? 

Great job all around, though! Such a nice PR 😄 👍 ❤️

src/OrbitDB.js
```diff
@@ -345,6 +345,11 @@ class OrbitDB {
     return Object.keys(databaseTypes).includes(type)
   }
 
+  static addType (type, store) {
+    if (databaseTypes[type]) throw new Error(`Type already exists: ${type}`)
+    databaseTypes[type] = store
```
haadcode: `databaseTypes` is defined on https://github.com/garbados/orbit-db/blob/1ef24dfa4be6884103c72e3762f163b8ae29a7a3/src/OrbitDB.js#L21 as `const`. Can you change that to be `let` to make sure the type describes it's a variable that will/can change?

test/create-type.test.js
```diff
@@ -0,0 +1,50 @@
...
+    if (ipfs) await ipfs.stop()
+  })
+
+  describe('addType', function () {
+    it('should have the correct custom type', async () => {
+      OrbitDB.addType(CustomStore.type, CustomStore)
+      let store = await orbitdb.create(dbPath, CustomStore.type)
+      assert.equal(store._type, CustomStore.type)
+    })
+  })
```
haadcode: Can we add a test that check that the existing database types can't be overwritten? (Essentially check the first line of `addType()`)

src/OrbitDB.js
```diff
@@ -345,6 +345,11 @@ class OrbitDB {
     return Object.keys(databaseTypes).includes(type)
   }
 
+  static addType (type, store) {
```
haadcode: Not gonna bikeshed this, but what do you think about naming it `addDatabaseType`? If you're happy with the current name, then let's go with that.

If you do change it, make sure to reflect that in the documentation too.

garbados:
OK, I folded in your feedback and squashed the commits. I hope it looks alright!

haadcode:
Looks awesome! 😄  Thank you for the PR @garbados and welcome to the contributors ❤️ 

I'll merge it an publish a new version on npm.

haadcode:
Released as [0.19.3](https://www.npmjs.com/package/orbit-db) on npm.

PR #322, opened 2018-02-08
Allow a custom keystore object to be specified via options

### LINKED ISSUE #319: Custom keystore
`orbit-db` needs a keystore to sign / verify signatures. 

Currently, we're using `orbit-db-keystore` (https://github.com/orbitdb/orbit-db/blob/master/src/OrbitDB.js#L38) but being able to specify a custom keystore to be used via config would be interesting for use-cases where there's already a keystore / signing functionality in the (d)app.

I can try to provide a PR that uses a custom keystore specified in `options` following the current API – if that's the best way to go. Thoughts?

thiagodelgado111:
This PR adds a check for `keystore` option, this way we can implement our own keystores or reuse it in cases where the app already have that. The keystore should follow `orbit-db-keystore`'s API since `orbit-db-store` and `ipfs-log` depends on it.

Closes #319

### DIFF
--- a/src/OrbitDB.js
+++ b/src/OrbitDB.js
@@ -35,7 +35,7 @@ class OrbitDB {
       : new Pubsub(this._ipfs, this.id)
     this.stores = {}
     this.directory = directory || './orbitdb'
-    this.keystore = Keystore.create(path.join(this.directory, this.id, '/keystore'))
+    this.keystore = options.keystore || Keystore.create(path.join(this.directory, this.id, '/keystore'))
     this.key = this.keystore.getKey(this.id) || this.keystore.createKey(this.id)
   }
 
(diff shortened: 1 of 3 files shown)


thiagodelgado111:
I've added a couple of tests but I don't know if that's enough, let me know if I should test different scenarios as well.

haadcode:
@thiagodelgado111 thanks for the PR! ❤️This is a great enhancement for the key side of things in OrbitDB. 

The keystore part hasn't received a lot of love yet, so I think this is fine as it is, but I reckon we'll revisit the keystore (at least the public API / interface) in the future. 

Going to merge!

thiagodelgado111:
Yay! Happy to help! :smile: :+1:

PR #325, opened 2018-02-21
Fix #311 - Avoid updating remote db address if it hasn't changed

### LINKED ISSUE #311: Remote database address on web demo is almost impossible to copy
This is a usability issue on Chrome, might not be on some other browsers. 

- Start the web demo
- Enter a database name and open for event log
- You'll see the continually updating results

screen shot 2018-01-21 at 08 40 19

- Try to select the event log address to open a remote database
- The JS keeps re-writing the DOM, including the element containing the address, making it almost impossible to time the copy action

It would be ideal to keep this particular element static and only update the ones below it, so that the user can more easily select it and test replication.

claus:


claus:
Issues like these make you appreciate libraries like React, even for the tiniest of apps, doesn't it? ;)

haadcode:
Thank you @claus for taking the initiative and fixing this! ❤️👍The fix looks good, will merge.

Indeed it does make you appreciate React ;)

Closes #311.

haadcode:
@claus I just realized this was your first official contribution to OrbitDB. Thank you and welcome to the contributors! 😄

*For the record, @claus has been contributing to projects related to OrbitDB for a long time and in various ways, but this was the first PR in this main repo and it warrants separate kudos :)*

claus:
Hahah, slow down, it was only two-or-so silly UI contributions. Wait for my next super massive PRs! 😁 Anyways, happy to help :)

Issue #273, opened 2017-12-07
Support ipfs-api again


haadcode:
For various reasons, we dropped the support for `ipfs-api` some months ago. We should bring it back.

One of the main reason to drop the support was the API mismatch between `ipfs` and `ipfs-api` especially when it comes to testing. The nodes are created in a different way and so this would require to re-factor the tests (mainly the before() parts where we start the IPFS nodes). There may be other dragons! 🐲 But from functional/technical perspective, it ***should*** work out of the box.

nmarley:
Wow, had no idea this wasn't supported any longer, but got bit by this also today. @haadcode What help do you need getting this to work again?

nmarley:
So it seems that with the write-permissioned database support, OrbitDB is looking like an entirely new product.

We've got a network of 4500 nodes and need to share IPFS peer ids in order to bootstrap the network connections. I had the idea to use OrbitDB _in order to share those_, but now with the addition of OrbitDB addresses, that's an additional layer which would have to be shared -- which is the point of us using OrbitDB in the first place, to avoid having to share unique values (the addresses) and add extra bootstrapping to our system. It's also now creating an "orbitdb" data directory, which was previously handled via ipfs (thru the ipfs-api).

Is it possible to restore these features from the previous version(s), or will this be the future of OrbitDB? I apologize for posting this here, but not sure where else to add design-related issues.

haadcode:
Thank you @nmarley for posting and raising your concerns. This is absolutely the right place to do that. And thank you for bringing this up, I had no idea you were using or planning to use OrbitDB at that scale.

It sounds to me this was an important part of your plan, so if you want to resolve it as fast as possible, please reach out to me on IRC or email me at haad@ipfs.io and I'd be more than happy to help.

We definitely would like to support your use case, so let's try break it down in order to figure out what's possible and what's needed:

> I had the idea to use OrbitDB in order to share those, but now with the addition of OrbitDB addresses, that's an additional layer which would have to be shared -- which is the point of us using OrbitDB in the first place, to avoid having to share unique values (the addresses)

Would you be able to elaborate on the mechanic of sharing the addresses? I don't understand why the OrbitDB addresses add an additional layer? If the addresses you intent to share are IPFS peer IDs, isn't that one database (thus one OrbitDB address)?

> It's also now creating an "orbitdb" data directory, which was previously handled via ipfs (thru the ipfs-api).

[...]

PR #328, opened 2018-02-27
Run tests with ipfs-api and go-ipfs again

haadcode:
This is a PR to make OrbitDB run tests with go-ipfs again (through js-ipfs-api) in addition to running them with js-ipfs.

This PR will:
- Remove obsolete/unused public API methods
- Run tests against both js-ipfs and go-ipfs
- Add write benchmark for using go-ipfs as the IPFS backend


diasdavid:
👏🏽👏🏽👏🏽👏🏽👏🏽

test/docstore.test.js
```diff
@@ -3,160 +3,169 @@
...
-      }
-      db = await orbitdb1.docstore(config.dbname, options)
+      if (ipfsd) 
+        await stopIpfs(ipfsd)
     })
 
-    afterEach(async () => {
-      await db.drop()
+    it('creates and opens a database', async () => {
+      db = await orbitdb1.docstore('first doc database')
```
thiagodelgado111: Duplicate?

haadcode: Good catch and it seems to be a problem with all store. They also don't seem to contain an assert for that test.

haadcode:
I'm going to split this PR in two, because there's really two major things in it atm:
- Use ipfsd-ctl to create test IPFS instances
- Run tests against both js-ipfs and go-ipfs

The first part is to just update the tests to use ipfsd-ctl and the second part brings in everything we need to run with go-ipfs again.

haadcode:
Closes #273

haadcode:
Using ipfsd-ctl to run the tests is now a separate PR here https://github.com/orbitdb/orbit-db/pull/335

haadcode:
This was now finally finished and in itself became a rather small PR compared to the original. The main work was done in the previous PRs (see commit log).

Going to merge this to master and keep testing it for some time before making a new release (0.20.0).

We now support go-ipfs again! 🎉 🚀 😄

PR #331, opened 2018-03-02
Update README.md

maxvisser:
helps people not wasting time on installing Orbit-db on windows when its not supported.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,8 @@ npm install orbit-db-cli -g
 
 ### As a library
 
+Orbit-db as a library is not supported on Windows.
+
 Install dependencies:
 
 ```


haadcode:
Thanks @maxvisser! 

How would you feel moving the notion of platform support under the "[Usage](https://github.com/maxvisser/orbit-db/blob/f462080c6e073b5443bedce3bb5feba2c8428061/README.md#usage)" title, one above the current location? I'm thinking it would fit better there and the fact that OrbitDB doesn't currently support Windows applies to the CLI, too. 

I would also propose to change the text slightly to: *"OrbitDB currently supports Linux and OS X, Windows is not supported yet."*

What do you think?

haadcode:
Thanks @maxvisser for the PR! 👍 Went ahead and merged it.

maxvisser:
I agree with your changes.  Thanks for merging it.

PR #358, opened 2018-04-13
Fix: tests weren't actually testing limit 0 cases

imrehg:


haadcode:
Great catch @imrehg! Thanks for the PR, much appreciated! ❤️👍

PR #360, opened 2018-04-14
Clarify eventlog and feed iterator options, and usage in tests

imrehg:
The `gt`/`gte`/`lt`/`lte` options for enventlog and feed iterators take the hash of an item, and that wasn't clear from the API docs. Also added more info to the `limit` option.

To see usage examples, many people are likely to turn to test suite. In the tests for eventlog and feed, the `items` variable first referred to an iterator with the actual items, and then later in the tests `items` was an array of the item hashes. This switch of types can be pretty confusing, thus the variable names were disambiguated.

### DIFF
--- a/API.md
+++ b/API.md
@@ -188,17 +188,19 @@ const db = await orbitdb.eventlog(anotherlogdb.address)
 
 **options** : It is an object which supports the following properties
 
-`gt - (string)`  Greater than
+`gt - (string)`  Greater than, takes an item's `hash`.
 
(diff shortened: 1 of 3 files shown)


haadcode:
What a nice improvement! ❤️Thank you @imrehg for the PR! These kind of improvements are hugely important for the developer experience.

As a side, I believe we called them items early on as we returned the full entry/op upon .add() but we don't do that anymore and just return the hash.

Thank you!

PR #361, opened 2018-04-17
Add Linux-specific installation details in README

vvp:
Instructions for running examples only applied to MacOS. Rather than introducing
a new NPM dependency for running platform-specific commands, I thought it's
simpler to just have own scripts for each platform.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,11 +127,24 @@ npm install --global babel-cli
 npm install --global webpack
 ```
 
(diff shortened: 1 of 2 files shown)


haadcode:
Thanks @vvp for the added details on how to run on Linux! Much appreciated! ❤️ 

Welcome to OrbitDB contributors! 😄👍

Issue #369, opened 2018-05-28
preparations for js-ipfs 0.29.0


daviddias:
Similar to https://github.com/orbitdb/ipfs-log/issues/124. Here I was less lucky, some tests failed with always the same error, example bellow:

```
 6) orbit-db - Automatic Replication (js-ipfs)
       automatic replication exchanges the correct heads:

      Uncaught AssertionError [ERR_ASSERTION]: must be a valid cid
      + expected - actual

      -false
      +true
```

haadcode:
Thanks @diasdavid for keeping us up to date! ❤️ 

Will try to work the new release into OrbitDB asap. Any ideas what the ^ `must be a valid cid` comes from? PubSub (as mentioned in https://github.com/orbitdb/ipfs-log/issues/124)?

mistakia:
It doesn't look like `orbit-db` or `ipfs-log` has `js-ipfs-block` in its dependency tree though.

diasdavid:
My recommendation is to update all deps.

PR #370, opened 2018-05-29
Updated (Dev) Deps

mistakia:
it seems updating `ipfs-repo` and `ipfsd-ctl` fixed the error in the tests mentioned in orbitdb/orbit-db#369


haadcode:
@mistakia thanks for creating the PR! ❤️ 

I created a PR to your fork in https://github.com/mistakia/orbit-db/pull/1#issue-191504216 which should fix the failing CI here.

Let me know if you need any help!

mistakia:
Looks like a different test has failed - I can look into it later tonight to see why they are flaky.

mistakia:
@haadcode I'm not sure what's going on with the `emits correct replication info on fresh replication` test on `go-ipfs`. 

It passed without changing anything. Earlier I moved the check to see if its finished [out from the `replicated` event](https://github.com/orbitdb/orbit-db/pull/370/commits/000f4cda41e8e5bbcd069d23f86935eef3280eaa#diff-b1d58ced77879cec2a109423fb3755a8L323) and [into the timer](https://github.com/orbitdb/orbit-db/pull/370/commits/000f4cda41e8e5bbcd069d23f86935eef3280eaa#diff-b1d58ced77879cec2a109423fb3755a8R324), in case the final `replicate.progress` event was called way after the `replicated` event. However, the same test still failed.

Please advise on how I should approach as I haven't been able to reproduce the failure locally.

mistakia:
I'm not sure if this is relevant yet but [`pubsub.unsubscribe` is async](https://github.com/ipfs/js-ipfs-api/blob/master/CHANGELOG.md#breaking-changes) but its not treated as such in [orbit-db-pubsub](https://github.com/orbitdb/orbit-db-pubsub/blob/master/src/ipfs-pubsub.js#L64)

mistakia:
In many of the after test hooks, `orbitdb.stop()` is called which then calls `orbitdb.disconnect()` then `IPFSPubsub.disconnect()` and thus `IPFSPubsub.unsubscribe()`.

I'm not certain if this is our race issue - because the same go-ipfs fails but this is one place where the behaviour is different between js-ipfs (sync unsubscribe) and go-ipfs (async unsubscribe).

mistakia:
I opened a new branch at [mistakia/orbit-db#test/unsubscribeAsync](https://github.com/mistakia/orbit-db/tree/test/unsubscribeAsync) with updated dev deps and fixes for unsubscribe. 

I'm going to run [the tests](https://circleci.com/gh/mistakia/orbit-db/tree/test%2FunsubscribeAsync) a number of times to try and catch the race condition... so far it's been passing.

mistakia:
It appears the race condition is gone on the [mistakia/orbit-db#test/unsubscribeAsync](https://github.com/mistakia/orbit-db/tree/test/unsubscribeAsync) branch. I'm going to close this pull request in favor of that one.

PR #371, opened 2018-05-30
Feature/access controller plus

coyotespike:
This PR is WIP, and completes the trio of PRs which start with `ipfs-log` and `orbit-db-store`.

This PR mainly (1) moves access controllers to their own repo, (2) accepts an access controller passed in via options, and (3) changes package.json to use the changed versions of `ipfs-log` and `orbit-db-store`. Therefore this is not a very important PR, mainly here to work with the other two.


coyotespike:
The dynamic access branch already allows us to switch access controller by type. And while it may make sense to move the access controllers to their own repo, that is not so important. Therefore closing this to concentrate on the dynamic access branch.

PR #384, opened 2018-06-14
Expose OrbitDBAddress isValid & parse

mistakia:


RichardLitt:
Thanks, @mistakia. Can you explain your rationale for doing this, a bit more?

mistakia:
Absolutely. 

I want to validate and reject malformed user inputed db addresses. Currently, if I were to just pass in the user input to the db constructor it will create a new db with the malformed address as the db name. But it looks like I can take care of that by just passing in `{ create: false }` so I don't have a need for this pr anymore.

RichardLitt:
@mistakia Thanks for explaining! Glad you resolved that. :)

mistakia:
Looks like I may have another need for this. 

I am building a [scuttlebot](http://scuttlebot.io/)-like log, where one of the [entry types](https://github.com/mistakia/record-node/blob/master/log/ContactEntry.js#L12), called a `contact`, is a pointer to another orbitdb log. I would like to valid addresses before users add a contact entry.

mistakia:
Also - instead of exposing them on an orbitdb instance. It may be preferable to expose them via `module.exports`.

RichardLitt:
Awesome. I want someone else to review this first (I don't know enough), but it certainly looks like a cool use case to me.

haadcode:
Thank you @mistakia for the PR! ❤️Makes sense to expose the address utilities so that addresses can be verified.

Also, awesome to see a test in the PR! 👍💯 

What do you think about exposing the functions as static class functions (like for example [isValidType](https://github.com/mistakia/orbit-db/blob/134ff233803707f40cf7406ff31ecf504fd8da22/src/OrbitDB.js#L370))? We would then call it with `OrbitDB.isValidAddress(...)` and `OrbitDB.parseAddress(...)`. The benefit would be that the user doesn't have to create the `orbitdb` instance before being able to use the functions. I feel this would be a bit nicer user experience.

Let us know and if you agree, would love to see the PR updated!

haadcode:
Thank you @mistakia for the PR! ❤️ This looks good now, will merge. Sorry it took so long to get back to to this. Slowly working through a backlog.

haadcode:
TODO:
- document the exposed API in API.md

PR #385, opened 2018-06-20
_replicationInfo no longer exists and breaks the browser example code.

strotter:
Changed to _replicationStatus to match tests and latest code and the example works again.


RichardLitt:
Nice catch!

haadcode:
Indeed great catch! Thank you @strotter for the PR ❤️

Issue #391, opened 2018-06-27
make clean leaves dirty package.json files


RichardLitt:
We should change `make clean` to not do this:

```sh
11:20 ~/src/orbitdb/orbit-db (master) * 🐕  make clean
rm -rf orbitdb/
rm -rf node_modules/
rm package-lock.json
11:20 ~/src/orbitdb/orbit-db (master) * 🐕  git k
* master
RichardLitt	git@github.com:RichardLitt/orbit-db.git (fetch)
RichardLitt	git@github.com:RichardLitt/orbit-db.git (push)
origin	git@github.com:orbitdb/orbit-db.git (fetch)
origin	git@github.com:orbitdb/orbit-db.git (push)
On branch master
Your branch is ahead of 'origin/master' by 38 commits.
  (use "git push" to publish your local commits)
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	deleted:    package-lock.json
	modified:   package.json

no changes added to commit (use "git add" and/or "git commit -a")
```

thiagodelgado111:
Tried to reproduce that from `master`. It did remove `package-lock.json` but it didn't modify `package.json`. BTW, should `make clean` remove that lockfile?

RichardLitt:
Did you run `npm install` first? 

`package-lock.json` should be checked in to source control, yeah. It shouldn't be removed.

thiagodelgado111:
Yep. You didn't install anything, just installed dependencies, right?

RichardLitt:
Yeah. Just tried again; the only thing `make clean` did that it shouldn't was delete the `package-lock.json` file, which it oughtn't to do.

thiagodelgado111:
I think it does the trick: https://github.com/orbitdb/orbit-db/pull/401/files ?

haadcode:
The intention for removing `package-lock.json` in `clean` was to provide a way to rebuild everything from scratch since package-lock can contain binary modules that are tied to  eg. OSX. By running `make clean && make` devs on Linux could get everything installed as their platform requirements.

This is very valuable functionality in the dev workflow. If we want to [change `clean`](https://github.com/orbitdb/orbit-db/pull/401) to **not** remove package-lock, let's make sure there's a command that replaces it. I propose moving the removal of package-lock to `make rebuild` command (and in `rebuild`, run `clean` as well) as then:
[...]

PR #401, opened 2018-07-10
Keep "package-lock.json" on make clean

thiagodelgado111:


RichardLitt:
Related to https://github.com/orbitdb/orbit-db/issues/391. Does the trick, I think.

haadcode:
As per https://github.com/orbitdb/orbit-db/issues/391#issuecomment-405920680, we should provide alternative command do achieve what `make clean` achieved before this PR.

I'm ok merging this if we have another command to replace the functionality.

haadcode:
Hey @thiagodelgado111 and @RichardLitt I went ahead and solved this by adding the relevant Makefile scripts and changes in https://github.com/orbitdb/orbit-db/pull/499.

Should do now what was requested in this PR but also address my concerns.

PR #417, opened 2018-07-18
Use go-ipfs 0.4.16

haadcode:
This PR will update the go-ipfs module (and ipfsd-ctl) to use version 0.4.16.

However, I'm starting to see weird behaviour locally and the tests don't pass. I don't know why. They work with 0.4.13. I won't have time to look into this, but leaving the PR open here if anyone wants to try to finish this and make the tests pass.


mistakia:
Did a little digging and it's caused by `waitForPeers`.

For instance, on the `counters -> sync counters` test: [`waitForPeers`](https://github.com/orbitdb/orbit-db/blob/master/test/counterdb.test.js#L121) polls for peers but [`ipfs.pubsub.peers`](https://github.com/orbitdb/orbit-db/blob/master/test/utils/wait-for-peers.js#L6) continues to return an empty array. Once the test timeout is hit, IPFS is shutdown while the interval continues to poll, which leads to the `Unhandled promise rejection`.

Part of the solution would be to make sure the interval is cleared once the test times out and fails. However, the problem is the instances are not peering.

I should have some time next week to take another look if the issue still exists.

haadcode:
I looked a little more into this: running the tests locally, they fail about 50% of the time. Sometimes `waitForPeers` (which calls `ipfs.pubsub.peers(topic)`) returns the other peer in the array and the test passes (as expected) and some times not and the test fails.

After adding ```const swarmPeers = await ipfs.swarm.peers()``` to [waitForPeers()](https://github.com/orbitdb/orbit-db/blob/master/test/utils/wait-for-peers.js#L7) as well as capturing the output of [ipfs.swarm.connect()](https://github.com/orbitdb/orbit-db/blob/master/test/utils/connect-peers.js), I can see the following output when the test fails:
```
swarm.connect1> { Strings:
   [ 'connect QmZ4uZZ9gm3dZCkr9dLUGxUaxxMyyiDcV1wEhHGLzx6fxJ success' ] }
swarm.connect2> { Strings:
   [ 'connect QmaYKELBbknQXiQSiWmDhKV4Rp73faV3jWen9cRb4xsjpA success' ] }
```

Which indicates connecting the peers was a success. Then later on, when querying the swarm.peers in waitForPeers:
```
pubsubTopicPeers> []
swarmPeers> [{ addr: <Multiaddr 04c0a8641506d5a4 - /ip4/192.168.100.21/tcp/54692>,
    peer:
     PeerId {
       _id: <Buffer 12 20 69 b1 d3 f3 d5 7e 3c 77 23 e1 bf 0f 0b 31 55 44 14 25 f2 4f 25 3d e9 95 e5 4d 7f 28 8a 7c ef 9e>,
       _idB58String: 'QmVTEnkvBzX2VEafmZbfyu9jTEZYxWybjj7RNj9TNhYdeu',
       _privKey: undefined,
       _pubKey: undefined },
    muxer: '' }]
```

Which indicates the peers are still connected on swarm level, but not on pubsub level.

Makes me think something got changed in pubsub/floodsub in 0.4.16 and we're seeing a regression bug.

haadcode:
Requires a fix for https://github.com/libp2p/go-libp2p-pubsub/issues/93.

haadcode:
Closing as obsolete. OrbitDB now uses go-ipfs v0.4.20. Should the problems described here persist, let's open new issues/PRs to handle them.

PR #422, opened 2018-07-25
docs: Add note about community calls

### LINKED ISSUE #394: Start a community call
Having talked to several people about this privately, it seems that there's interest to do a bi-/weekly community call for OrbitDB to discuss progress, ask questions and general get together.

Let's try to make it a fixed time every week taking into consideration All The Timezones, which I suppose means sometime between 3pm to 5pm or similar. Another thing to consider is that many OS teams seem to have weekly calls on Monday's and people might have scheduling conflicts, so we should be considerate of that too.

Would love to hear who's interested to join on (semi-)regular basis and what the preferred times are?

Would Monday be a good day, or should we consider Tuesday or Wednesday?

Should we do a weekly or bi-weekly call?

RichardLitt:
This should close #394, as we have moved the calls issues to that repository.


haadcode:
LGTM 👍

PR #426, opened 2018-08-06
docs(api): fix broken links

mistakia:


RichardLitt:
Thanks, @mistakia!

Did you use an automated tool to do this? Could we add this as a process so that this sort of PR isn't needed again?

mistakia:
I did not - found it while checking the docs. That would be a handy tool 👍 

However, I have noticed that when you're editing markdown on github and hit "preview" - broken links will show up "black" with no link styling.

mistakia:
Closing in lieu of #428

PR #433, opened 2018-08-09
Wait for send to complete on exchange heads

crazybuster:
For ipfs-api, during heavy load when multiple db heads are being exchanged if there is no wait here for the send to complete then the actually publish won't complete and the peers would not receive the heads sent to it. This seems to alleviate the problem, though it's not totally predictable.


haadcode:
@crazybuster thanks for the PR! 👍 Looks good to me and will pull it in. Small PRs like this are great! 😄 

I believe the "not always sending heads" problem is actually a case of peers not connecting on Pubsub-level, or at least that's the behaviour we've witnessed in OrbitDB tests. This, I believe, is a regression bug in IPFS/libp2p, see https://github.com/orbitdb/orbit-db/pull/417#issuecomment-407002661 and https://github.com/libp2p/go-floodsub/issues/93 for more details. However, if you're seeing different kind of "not always sending the heads", please let us know!

Issue #425, opened 2018-08-06
Using orbit-db without pubsub


oed:
Is it possible to use orbit-db with a go-ipfs node that doesn't have pubsub enabled? (My guess is no, not at this point) How hard would it be to make this happen?

The use case I'm considering is the following. User creates a db locally (or connected to a remote go-ipfs node). It then sends the hash of the db to a server which fetches and pins all the objects of the db. When the user goes online with a different computer they can ask the server about the latest hash and fetch the data. This could also be done similarly with multiple users of a db.

mistakia:
Doable and shouldn't be too hard — depends more on content discovery (i.e. DHT, etc) working well on IPFS.

The server will need the head entry. Using the head, the server can retrieve the previous entries by following the `next` pointer and feeding them to [`Log.fromEntryHash`](https://github.com/orbitdb/ipfs-log/blob/master/src/log.js#L446) — the entries will need to be available and discoverable on IPFS.

A few good spots to look under the hood: [`Store.sync()`](https://github.com/orbitdb/orbit-db-store/blob/master/src/Store.js#L222) & [`Replicator._processOne()`](https://github.com/orbitdb/orbit-db-store/blob/master/src/Replicator.js#L152)

mistakia:
If my understanding is accurate — Pubsub is used to efficiently exchange heads between peers of a given db. Without Pubsub, you'll have to handle the exchange of heads oob but can then feed the heads to `Store.sync()`.

oed:
Thanks, sounds like it shouldn't be too hard to implement. My main problem thought is that when creating a db on an ipfs node without pubsub enabled I get this error:
```
UnhandledPromiseRejectionWarning: Error: experimental pubsub feature not enabled. Run daemon with --enable-pubsub-experiment to use
```

mistakia:
Hmm - my first suggestion would be to pass `{replicate: false}` as an option when creating

oed:
That worked, thanks! Didn't know it was an option, don't think it's in the API doc :)

PR #436, opened 2018-08-16
Document replicate option

mistakia:
Add missing `replicate` config option for `OrbitDB.create()` and `OrbitDB.open()`.

Related #425

### DIFF
--- a/API.md
+++ b/API.md
@@ -92,6 +92,7 @@ Returns a `Promise` that resolves to [a database instance](#store-api). `name` (
 - `directory` (string): The directory where data will be stored (Default: uses directory option passed to OrbitDB constructor or `./orbitdb` if none was provided).
 - `write` (array): An array of hex encoded public keys which are used to set write acces to the database. `["*"]` can be passed in to give write access to everyone. See the [GETTING STARTED](https://github.com/orbitdb/orbit-db/blob/master/GUIDE.md) guide for more info. (Default: uses the OrbitDB instance key `orbitdb.key`, which would give write access only to yourself)
 - `overwrite` (boolean): Overwrite an existing database (Default: `false`)
+- `replicate` (boolean): Replicate the database with peers, requires IPFS PubSub. (Default: `true`)
 ```javascript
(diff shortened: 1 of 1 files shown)


API.md
```diff
@@ -92,6 +92,7 @@ Returns a `Promise` that resolves to [a database instance](#store-api). `name` (
 - `directory` (string): The directory where data will be stored (Default: uses directory option passed to OrbitDB constructor or `./orbitdb` if none was provided).
 - `write` (array): An array of hex encoded public keys which are used to set write acces to the database. `["*"]` can be passed in to give write access to everyone. See the [GETTING STARTED](https://github.com/orbitdb/orbit-db/blob/master/GUIDE.md) guide for more info. (Default: uses the OrbitDB instance key `orbitdb.key`, which would give write access only to yourself)
 - `overwrite` (boolean): Overwrite an existing database (Default: `false`)
+- `replicate` (boolean): Subscribe & exchange heads over IPFS PubSub. (Default: `true`)
```
haadcode: I think a better description for this is *"Replicate the database with peers."* as that's what it does atm (see https://github.com/orbitdb/orbit-db/blob/master/src/OrbitDB.js#L146). If set to `false`, OrbitDB won't subscribe to the pubsub topic and thus will not receive any updates from peers nor send updates to peers.

haadcode:
Thank you @mistakia 👍Left a small request to change the description, otherwise LGTM.

API.md
mistakia: How about "Replicate the database with peers over IPFS PubSub"?

haadcode: Sounds good 👍

haadcode: (although technically the *replication* doesn't happen over pubsub 😉but it's fine to describe it as such)

mistakia: ah right - just the exchange of heads. I'll switch to "Replicate the database with peers (requires IPFS PubSub)"?

Just want to inform those who (strangely) are not using PubSub.

haadcode: 👍

haadcode:
Thanks, good to merge now!

PR #443, opened 2018-08-31
Dockerfile for OrbitDB base image

vvp:
This PR defines the base image for running OrbitDB code in containers. It's something that I've personally been using for a while, so it can already be used locally for running tests or examples in isolated container. Besides the initial Dockerfile for building OrbitDB base images, there's some instructions how to use it with Docker. 

There is optimization left to do in base image itself to make it leaner (smaller nodejs-base image, multi-stage builds etc). Also I've only been using it on Linux where there are usually no problems with containers, so it'd be great to hear experiences from Docker for Mac/Windows users.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,26 @@
+# Running OrbitDB code in containers
+
+[Docker](https://github.com/docker/docker-ce) is a tool for running software in containers ie. OS-level virtualization. This directory contains the needed configuration files to:
+
+- Build base Docker images for OrbitDB
+
+
+## Base Docker image
+
+[Dockerfile](Dockerfile) defines OrbitDB base images that are based on [official node.js -image](https://hub.docker.com/_/node/). 
+
+Build local images with command (in repository root):
+
+```bash
+docker build -t orbit-db -f docker/Dockerfile .
+```
+
+After building local image, run node.js-examples inside container:
+
+```bash
+docker run -ti --rm orbit-db npm run examples:node
+```
+
+## Tested versions
+
+- Docker 1.13.1 (Fedora Linux 27)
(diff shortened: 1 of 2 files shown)


haadcode:
Looks good to me but I'm not very familiar with Docker.

thiagodelgado111:
I have (almost) no experience with Docker. If anything, I've used it to provide things I needed to work, like: database, email server, ipfs, etc. On the Dockerfile, the only thing that occurred to me is that maybe `package-lock.json` should be copied too? I was gonna mention using `npm ci` but I think it probably isn't relevant for the case

docker/README.md
```diff
@@ -0,0 +1,35 @@
...
+After building local image, run node.js-examples inside container:
+
+```bash
+docker run -ti --rm orbit-db npm run examples:node
+```
+
+## Why would you want to run OrbitDB in container?
+
+Containers are nice because as software execution environments they are:
+- Reproducible, which helps testing and development because you can revert container to original state by destroying it and creating it again,
```
RichardLitt: revert the container

docker/README.md
```diff
@@ -0,0 +1,35 @@
...
+- Isolated, which guarantees that external factors like npm versions, operating system version, or other installed software like native compilers do not affect the execution.
```
RichardLitt: npm versions, the operating system version, or..

vvp:
@RichardLitt Thanks for proof-reading, fixed! :+1: 

@thiagodelgado111 good thing in not copying that lock-file into image is that it lets you build multiple local Docker images from the same OrbitDB package.json but with different dependencies. With them it's easy to setup a continuous testing for catching breaking dependencies as soon as possible (given that npm install or npm test catches them :slightly_smiling_face: )

haadcode:
Thanks for PR and the docker support @vvp! 👍❤️

PR #449, opened 2018-09-11
Feat/new ipfs log (wip)

shamb0t:
This PR updates orbit-db to use the lastest ipfs-log

### TODO
- [x] Discuss style / init methods
- [ ] Fix dependencies
- [ ] Update docs

Depends on on https://github.com/orbitdb/orbit-db-store/pull/33


haadcode:
Looks a lot better now without the `initialize` work around.

I wonder though, is `init` the best name for the factory (it was proposed by me)? I believe @thiagodelgado111 proposed `connect` but I feel that's also not quite what we should call it. What do you think of `start` (as in "start an OrbitDB instance")?

thiagodelgado111:
We talked about going with `create` :)

haadcode:
Right, I remember `create` was mentioned. I feel it may get confuse with `orbitdb.create` instance method (which I'm unsure we should keep anyways, might be better to just have `open` and `create: true` as an option but that's another story).

My intuition in proposing `start` is somewhat related to IPFS and while IPFS doesn't require to call start explicitly, it does require to *listen for `start`* event.

src/OrbitDB.js
```diff
@@ -30,17 +31,26 @@ let databaseTypes = {
...
     this.directory = directory || './orbitdb'
-    this.keystore = options.keystore || Keystore.create(path.join(this.directory, this.id, '/keystore'))
-    this.key = this.keystore.getKey(this.id) || this.keystore.createKey(this.id)
+    this.stores = {}
     this._directConnections = {}
   }
 
+  static async init (ipfs, directory, options = {}) {
+    const { id } = await ipfs.id()
+    directory = directory || './orbitdb'
```
haadcode: We could move `directory` to options here, right? If there's no directory provided, we have a default for it and as such, it could be consider *optional*. What do you think? (this would give us the opportunity to eventually move it to optional arg in the constructor, too, in subsequent refactoring)

shamb0t: Yea agree we should it to options

src/OrbitDB.js
```diff
@@ -30,17 +31,26 @@ let databaseTypes = {
 class OrbitDB {
   constructor(ipfs, directory, options = {}) {
     this._ipfs = ipfs
-    this.id = options.peerId || (this._ipfs._peerInfo ? this._ipfs._peerInfo.id._idB58String : 'default')
+    this.id = options.peerId
+    this.identity = options.identity
```
haadcode: Should identity be a required arg? I believe nothing on the log-level will work if there's no identity.

haadcode:
One small comment on requiring `directory` arg and the question of naming `init`, but otherwise looks good 👍

src/OrbitDB.js
shamb0t: Was trying to keep the interface the same, but if we are going to break things then yes I agree we should require identity.

haadcode: Agreed, we should try to not break things now and can leave this for later refactoring (knowing we'll break things then).

test/write-permissions.test.js
```diff
@@ -70,15 +70,21 @@ Object.keys(testAPIs).forEach(API => {
       rmrf.sync(dbPath)
       ipfsd = await startIpfs(API, config.daemon1)
       ipfs = ipfsd.api
-      orbitdb1 = new OrbitDB(ipfs, dbPath + '/1')
-      orbitdb2 = new OrbitDB(ipfs, dbPath + '/2')
+      orbitdb1 = await OrbitDB.createInstance(ipfs, {
+        directory: dbPath + '/1',
+        id: 'A'
```
haadcode: Do we need to specify the `id` here? If not, we could remove these to simplify the code further.

test/write-permissions.test.js
```diff
@@ -198,8 +202,9 @@ Object.keys(testAPIs).forEach(API => {
             await database.tryInsert(db2)
           } catch (e) {
             // Make sure peer 2's instance throws an error
-            assert.equal(e.toString(), 'Error: Not allowed to write')
+            err = e.toString()
```
haadcode: Good catch 👍

src/OrbitDB.js
```diff
@@ -30,17 +31,26 @@ let databaseTypes = {
...
       : new Pubsub(this._ipfs, this.id)
-    this.stores = {}
     this.directory = directory || './orbitdb'
-    this.keystore = options.keystore || Keystore.create(path.join(this.directory, this.id, '/keystore'))
-    this.key = this.keystore.getKey(this.id) || this.keystore.createKey(this.id)
+    this.stores = {}
     this._directConnections = {}
   }
 
+  static async createInstance (ipfs, options = {}) {
```
haadcode: Since we're changing the way OrbitDB instances should be created (for the better 👍), we should also update the docs right away. We don't necessarily have to do it in this PR, but should at least leave a TODO to do it immediately after.

test/write-permissions.test.js
shamb0t: Keystore makes an identity using the ipfs peer id, since they are the same the identities would be the same here, the optional id is to use a different id in identity creation. I would rather find a better way but this was quick

haadcode: Got it 👍 Let's go with this then.

haadcode:
Looks good to me to merge 👍Are there anymore expected updates? @thiagodelgado111 let us know if this looks good to you.

src/OrbitDB.js
```diff
@@ -28,26 +28,26 @@ let databaseTypes = {
...
 }
 
-class OrbitDB {
-  constructor(ipfs, directory, options = {}) {
+  class OrbitDB {
+  constructor(ipfs, identity, options = {}) {
     this._ipfs = ipfs
     this.id = options.peerId
-    this.identity = options.identity
+    this.identity = identity
```
haadcode: Should we throw an error here if identity wasn't provided since it's required (I know we don't do that for ipfs yet either, but would prolly be good to throw to guide the user to call the constructor correctly)?

haadcode: If we do ^, could also add the same check for IPFS while we're at it.

API.md
```diff
@@ -63,19 +65,35 @@ Read the **[GETTING STARTED](https://github.com/orbitdb/orbit-db/blob/master/GUI
...
+- `directory` (string): path to be used for the database files. By default it uses `'./orbitdb'`.
 
 - `peerId` (string): By default it uses the base58 string of the ipfs peer id.
 
 - `keystore` (Keystore Instance) : By default creates an instance of [Keystore](https://github.com/orbitdb/orbit-db-keystore). A custom keystore instance can be used, see [this](https://github.com/orbitdb/orbit-db/blob/master/test/utils/custom-test-keystore.js) for an example.
 
-After creating an `OrbitDB` instance , you can access the different data stores. Creating a database instance, eg. with `orbitdb.keyvalue(...)`, returns a *Promise* that resolves to a [database instance](#store-api). See the [Store](#store-api) section for details of common methods and properties.
+- `identity` (Identity Instance): By default it creates an instance of [Identity](https://github.com/orbitdb/orbit-db-identity-provider/src/identity.js)
+
+### constructor(ipfs, identity, [options])
```
haadcode: I was thinking if we should leave out documenting the constructor. I believe what we want the users to do is to use `createInstance` instead of the constructor. If we document both, the user might think either one can be used and would prolly default to using the familiar one, the constructor.

Thoughts?

API.md
```diff
@@ -175,6 +193,13 @@ Returns an `Object` with the contents of the entry.
   // { name: 'Friend' }
   ```
 
+#### del(key)
```
haadcode: 👍 

Closes https://github.com/orbitdb/orbit-db/issues/452

src/OrbitDB.js
```diff
@@ -27,20 +28,38 @@ let databaseTypes = {
   'keyvalue': KeyValueStore,
 }
 
-class OrbitDB {
-  constructor(ipfs, directory, options = {}) {
+  class OrbitDB {
+  constructor(ipfs, identity, options = {}) {
+    if (!isDefined(ipfs))
```
haadcode: 👍

src/OrbitDB.js
```diff
@@ -27,20 +28,38 @@ let databaseTypes = {
...
+      throw new Error('IPFS is a required argument. See https://github.com/orbitdb/orbit-db/blob/master/API.md#createinstance')
+
+    if (!isDefined(identity))
```
haadcode: 👍

API.md
shamb0t: I was thinking the same but thought perhaps we wanted them to use the constructor instead....this would however force users to create an identity first instead of having a default one created for them.

shamb0t: Actually maybe it is better to have a default created since that should be what is used out of the box. Advanced users can create their own identities and pass it as an option. Agreed its more confusing to have another way to create an instance.

src/OrbitDB.js
```diff
@@ -9,11 +9,12 @@ const DocumentStore = require('orbit-db-docstore')
 const Pubsub = require('orbit-db-pubsub')
 const Cache = require('orbit-db-cache')
 const Keystore = require('orbit-db-keystore')
+const IdentityProvider = require('orbit-db-identity-provider')
```
haadcode: After some initial user tests and fresh experiences on this, I realized that we should expose the IdentityProvider to the users so that they can actually set up custom identities :) Ie. export it like we do in ipfs-log.

thiagodelgado111:
Hi there, @haadcode and @shamb0t, I've put up a branch on based on top of this one with a few changes, especially related to how the access controller/identity is passed to the store. Of course, it doesn't look right but let's discuss what's the best way to do it :) 

https://github.com/shamb0t/orbit-db/compare/feat/new-ipfs-log...thiagodelgado111:maintenance/acl-update

vongohren:
I have to say it works well for me, the identity part here.
I might be doing it a bit to complicated, but works for my use case.
When is this work progressing to merge?

src/OrbitDB.js
haadcode: Take that back. User can do `const IdentityProvider = require('orbit-db-identity-provider')` whenever they use `orbit-db` since orbitdb pulls in `orbit-db-identity-provider`.

vongohren:
> Hi there, @haadcode and @shamb0t, I've put up a branch on based on top of this one with a few changes, especially related to how the access controller/identity is passed to the store. Of course, it doesn't look right but let's discuss what's the best way to do it :)
> 
> [shamb0t/orbit-db@feat/new-ipfs-log...thiagodelgado111:maintenance/acl-update](https://github.com/shamb0t/orbit-db/compare/feat/new-ipfs-log...thiagodelgado111:maintenance/acl-update)

Can any of these extra changes be related to my problems with using this branch @haadcode @thiagodelgado111. My main problem is that I cant seem to get the subpub peers to recieve my data, so my frontend is not able to read the data that some backend is writing to a particular adresss

shamb0t:
Superseded by #495

PR #455, opened 2018-09-16
fixed specified

0xflotus:


haadcode:
Thanks for the PR @0xflotus, good catch!

RichardLitt:
Thanks for your first PR, @0xflotus. :)

PR #459, opened 2018-09-17
docs(api): fix outdated doc

haoliangyu:
In the latest docstore code, the docstore.get() function does not use the `fullOp` option to retrieve a document ([code](https://github.com/orbitdb/orbit-db-docstore/blob/master/src/DocumentStore.js#L30)). Therefore, the docstore.get() function should by default return the payload value ([code](https://github.com/orbitdb/orbit-db-docstore/blob/master/src/DocumentIndex.js#L11)) and there is no need to add the secondary mapper function any more.


shamb0t:
Thanks @haoliangyu!

PR #489, opened 2018-10-30
Make manifest/db address verifiable

### LINKED ISSUE #488: Make orbit-db manifest/address verifiable
When we get an `orbit-db` address, we need to make sure that that is really the manifest created by the entity we're "assuming" that created it. So one possible solution for it is to add a "signature" as part of the database address so it'd look like this: `/orbitdb/<manifest hash>/<db-name>/<manifest signature>`

That way we can simply load the manifest hash and verify the data on it against the manifest signature part of the address. For that, manifest creation would have to be changed and `orbit-db` addresses too.

As an example, this is what I think the db manifest object could contain:
```
- name: 'my-db'
- type: 'kvstore'
- controller: 'ethereum-wallet/0x6320'  | 'ipfs/QmP0abcd' | 'ethereum-contract/0x1234' | 'orbitdb/<manifest>/<name>/<signature>'
```

(In fact, we could add a lot of metadata there)

thiagodelgado111:
## Description

(This is based on `thiagodelgado111/maintenance/acl-update` and `shamb0t/orbit-db/feat/new-ipfs-log` branches)

Introduces a signature to orbit-db addresses to make manifests verifiable (via a contract signing them or a key)

Resolves #488


haadcode:
Thanks @thiagodelgado111 for the PR 👍 

I would like to understand the motivations behind adding the signature. The title of the PR is a bit misleading as ***we have verifiability*** of the manifest through the merkle DAG that the manifest and its references form. I assume you mean to "authenticity" with the added signature?

Generally, I don't think adding another hash/sig to the address itself is the best solution: it adds complexity to the address but would also break *all* current addresses. Changing the address scheme should only be done for very heavy reasons. Instead, these kind of things could be added to the manifest directly.

Second, going back to the motivations, I see many cases where we shouldn't have a signature for the manifest (bearing in mind it's still verifiable as a merkle dag) as that can't be changed at a later point and sometimes shouldn't even be tied to a particular key.

@thiagodelgado111 perhaps you could explain your use case or what you're trying to achieve in order to understand this better?

I'd like to discuss all of ^ before we consider merging this PR.

thiagodelgado111:
Hi @haadcode, thanks for taking the time to look at it. I'm trying to make sure the manifest was indeed created and signed by the contract or key I trust.

e.g: If I have a contract that will do AC for an db, I wanna make sure that that address and manifest were in fact signed/generated by the contract

aphelionz:
Closing this as it's out of date. Happy to re-open if/when necessary. Thank you!

PR #500, opened 2018-11-13
Determine address

tyleryasaka:
This change would allow an orbit-db address to be determined through the API, without actually creating the database. This is useful for applications where peers may wish to determine other peers' orbit-db addresses purely based on their public keys. (I implemented this while experimenting with [3box](https://github.com/uport-project/3box-js) and mentioned this to @oed).

~Currently this *does* produce side effects - the manifest and access controller are saved to IPFS. I don't currently see a way to get the IPFS hash from [`object.put`](https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/OBJECT.md#javascript---ipfsobjectputobj-options-callback) without actually writing the data, but would love to know if there's a way to do this.~

EDIT: this no longer produces side effects

I did add a couple of tests to the `create-open` file (not sure if that's the right place). I'm happy to update the docs if this is something that you want to consider merging.


joshfraser:
This would be very useful for us and the way we're using OrbitDB at Origin Protocol.

aphelionz:
@tyleryasaka Thanks for this! It appears that `ipfs add` has an `only-hash` option that might allow for what you're after. Try that in `js-ipfs` with:

```
ipfs.add(content, { "onlyHash": true })
```

Might that help remove the necessary side effect?

tyleryasaka:
@aphelionz Yeah, unfortunately we can't just use `ipfs add` here. We're using [`object.put`](https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/OBJECT.md#javascript---ipfsobjectputobj-options-callback) which does not have the `onlyHash` option currently. I started looking into the implementation of `object.put` to see if there's an easy way to see how they're building the hash, but it got hairy pretty quickly.

aphelionz:
@tyleryasaka ah, my mistake - I assume you've looked into this as well but just want to make sure... anything useful here? https://github.com/ipfs/js-ipfs/issues/1205#issuecomment-371134571

tyleryasaka:
@aphelionz I skimmed over that but must have missed that. Trying that method now. 🤞

tyleryasaka:
@aphelionz it worked! Thanks. 🎉

src/OrbitDB.js
```diff
@@ -244,13 +228,34 @@ class OrbitDB {
       accessController.add('write', this.key.getPublic('hex'))
```
oed: Would be nice if you can specify which keys to use here. If I'm not mistaken you can only determine the address on your own stores if you can't specify the public key(s).

tyleryasaka: @oed Yeah, that's true; really I just kept this here so that the functionality of `create` remains the same. If you're trying to determine someone else's address, you'd need to pass in a write option, like `{ write: [...] }`. I think that should be sufficient?

oed: Oh, sorry. Didn't read the code well enough! Please disregard the previous message :)

src/db-manifest.js
```diff
@@ -1,13 +1,22 @@
...
     name: name,
     type: type,
     accessController: path.join('/ipfs', accessControllerAddress),
   }
-  const dag = await ipfs.object.put(Buffer.from(JSON.stringify(manifest)))
+  let dag
+  const manifestJSON = JSON.stringify(manifest)
+  if (onlyHash) {
+    dag = await new Promise(resolve => {
+      DAGNode.create(Buffer.from(manifestJSON), (err, n) => { resolve(n) })
```
haadcode: I'd like to understand what the return values are here? Does  `DAGNode.create(...)` return the same hash as `ipfs.object.put(...)`?

haadcode: We should handle the possible `err` in the callback here. Maybe throw it (which would be caught by the surrounding try/catch?

haadcode:
Thanks @tyleryasaka for the PR! Agreed this would be a useful function to add 👍 I left couple of questions in the comments.

src/db-manifest.js
tyleryasaka: >I'd like to understand what the return values are here? Does DAGNode.create(...) return the same hash as ipfs.object.put(...)?

@haadcode correct, `DAGNode.create(...)` returns the same hash as `ipfs.object.put(...)`. I'm not clear on the specifics of the method, but I followed the example [here](https://github.com/ipfs/js-ipfs/issues/1205#issuecomment-371134571) that @aphelionz was kind enough to point me to.

>We should handle the possible err in the callback here. Maybe throw it (which would be caught by the surrounding try/catch?

Done 👍

oed:
Any updates on this? I'd like to use this feature fairly soon :)

aphelionz:
Pinging @haadcode and @oed for another look :)

haadcode:
Looks fine now functionality-wise 👍 There are few things that could be improved in the code itself (splitting code into more functions etc.), but I don't think that's necessary and we can refactor things later as we plan to do anyways (if you want to further improve it, let me know and I can give provide more detailed feedback). Thanks @tyleryasaka for your work on the PR! ❤️ 

Should resolve the conflicts before merging, but LGTM 👍

PR #507, opened 2018-11-21
feat: allow local cache to be pass as option

zachferland:
This PR adds two lines to allow a cache instance to be passed as an option to the `OrbitDB` constructor. While still defaulting to [orbit-db-cache](https://github.com/orbitdb/orbit-db-cache). 

At [3box](https://github.com/uport-project/3box-js) this change would help us implement this [PR](https://github.com/uport-project/3box-js/pull/185) where we are using orbitdb, but running both ipfs and a cache instance in an iframe, so that stores/data can be shared across domains. We are passing a proxy instance of both the cache and ipfs with [orbit-db-cache-postmsg-proxy](https://github.com/3box/orbit-db-cache-postmsg-proxy) and [ipfs-postmsg-proxy](https://github.com/ipfs-shipyard/ipfs-postmsg-proxy) to the `OrbitDb` constructor.


aphelionz:
@zachferland Thanks for this. We reviewed it on the community call and it looks good. Would you mind appending the test suite with a simple test for this functionality?

zachferland:
great, thanks @aphelionz, I can add some coverage for it

zachferland:
Adding test coverage by adding:
1) simple example custom cache
2) adding another test file to just run with all the stores  (maybe not needed, as 3 seems to cover everything)
3) run persistency.js with custom cache (as well as with default cache as it did before)

Added one more commit to move redundant test configs 

Let me know if any of the tests are redundant or not necessary, was not much test code added, but more than a simple test

RichardLitt:
Thank you, @zachferland! 🎉

PR #511, opened 2018-11-26
Add an FAQ document

RichardLitt:
This pull request adds an FAQ document. @haadcode has started opening issues for questions and labelling them as 'FAQ', which I think is a good start. However, I don't think that the questions should stay there. Issues which are open for a long time tend to inflate the issue count, which can negatively affect maintainer guilt around too many open issues, and can be confusing for new contributors who are looking for answers to specific questions and see too many issues with the keywords they're looking for. Having a document doesn't really solve this problem - documents grow stale, too - but I think it makes it easier for people to find the relevant questions, and more importantly it makes it much easier for maintainers to point to a canonical answer instead of answering the same question over and over again. It also obviates the issue of irrelevant drive-by comments.

I am not sure if a Markdown file is the best answer to the FAQ problem, but it's an answer we can implement and test _right now_, as opposed to say opening up a Discourse forum and then waiting a couple of months as people gradually migrate over there. The FAQ file isn't meant to be a permanent solution; but I want to test it. 

Obviously, there are more questions that could be in here, and these questions and answers could be improved. I've gone through and added some as a way of jumpstarting this. Please feel free to comment on the answers, here, or in the issues they were posted on. But let's focus on getting this document improved and merged, rather than nitpicking eternally about how a question is best answered. And, please, if you see questions that you answer _more than once_, put it in here. That includes _"When are the community calls?"_ and other questions, too.

What do you think?

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -5,3 +5,7 @@ OrbitDB, like all code, is in a state of constant development. Doubtless, you're
 This is a living document. If you see an answer that could be improved, please [open an issue](https://github.com/orbitdb/orbit-db/issues/new) or submit a PR directly. If you think than a question is missing, please [open an issue](https://github.com/orbitdb/orbit-db/issues/new). If you think that there is a better way to resolve a question - perhaps by improving the  `orbitdb --help` docs or by adding a feature - please [open an issue](https://github.com/orbitdb/orbit-db/issues/new). Sense a theme yet? :)
 
 This document is seeded by questions from people opening issues in this repository. If enough people ask the same question, we'll add one here and point newcomers to it. Please don't be offended if the maintainers say "read the FAQ" - it's our way of making sure we don't spend all of our time answering the same questions.
+
+## Can I recreate the entire database on another machine based on the address?
+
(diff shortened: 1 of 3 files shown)


RichardLitt:
Ping @orbitdb/maintainers.

haadcode:
Looks great! Thank you @RichardLitt for adding the doc 👍 

Is there anything else to add to this? It look ready to merge from my perspective.

RichardLitt:
Well, it's more a question of *Is this a good base*? Ideally, we could build on it. If the answers in here look right, then this is probably ready to merge, but I want to ensure that we start using it by adding answers and referring to it in the future. 

I also wanted feedback from others, as ideally this will cut down on all of our time. @aphelionz What do you think?

aphelionz:
If there was a way we could get people in the community writ large to lean on this document heavily and even get them to "ask" questions via PRs against this md file, that would be ideal IMO. The biggest challenge here is simply going to be remembering to use it - but I think we should try. Pin it to every piece of chat software and put it right near the top of the README file and :crossed_fingers:

Issue #509, opened 2018-11-22
Emit an event when database has finished syncing with connected peers


tyleryasaka:
https://github.com/orbitdb/orbit-db/issues/474 describes a situation in which this event would be useful. Existing events do not communicate this information.

The `load` event is emitted simply once a database has finished loading from memory. This event may be emitted before the database has synced with connected peers.

The `replicated` event is in fact emitted after a database has synced; however, it does not always fire (and there is no straightforward way to determine whether it will fire). For example, if a connected peer does not have the database open, or it does not have any heads for the database, then no replication will occur and the `replicated` event will not fire. Thus, it's uncertain whether the `replicated` event will ever fire, so one cannot reliably wait for this event to fire in order to know that a database has finished syncing with peers.

I have implemented a solution for this problem; however, this solution is not sufficient in its current form. I'm linking to it here as an example of what a solution might look like. The solution spans 2 repos:

[orbit-db changes](https://github.com/orbitdb/orbit-db/compare/master...projectaspen:synced-event)
[orbit-db-store changes](https://github.com/orbitdb/orbit-db-store/compare/master...projectaspen:wait-process-queue)

Further, here is an example of code that uses this new `synced` event to know when a database has finished syncing (usage is still a bit complex, as the event fires only once per peer/database combination, and requires additional pubsub message sending to request that a connected peer open a specific databse): https://github.com/projectaspen/aspen/blob/master/packages/orbit-connect-client/src/orbit-connect-client.js#L104

haadcode:
@tyleryasaka we discussed this yesterday in the community call. See [the notes here](https://github.com/orbitdb/welcome/blob/master/meeting-notes/2018/2018.11.27.md) (under "semantics of synced"). The notes should capture the (design/ux) challenge here, but if they don't, let us know and we can continue the discussion here.

tyleryasaka:
I've created a PR that implements the solution I am referring to. It makes minimal code changes, and does not touch the enqueueing system (unlike my original solution). It requires an accompanying change to orbit-db-store.

https://github.com/orbitdb/orbit-db/pull/514
https://github.com/orbitdb/orbit-db-store/pull/38

tyleryasaka:
[...]

PR #514, opened 2018-11-29
Add event peer.exchanged

tyleryasaka:
Requires: https://github.com/orbitdb/orbit-db-store/pull/38

This adds a new event:

```javascript
db.events.on('peer.exchanged', (peer, address, heads) => ... )
```

>Emitted after heads have been exchanged with a peer for a specific database. This will be emitted after every exchange, even if no heads are received from the peer, or if all received heads are already present. (This is in contrast with the `replicated` event, which will only fire when new heads have been received.) Note that `heads` here contains heads *received* as part of the exchange, not heads sent.

I'm open to a better event name.

This offers 2 things that the `replicated` event does not:
1. It should always fire when peers exchange heads for a database (and only after all of the sent heads have been synced). The `replicated` event does not fire if there are no heads sent, or if all heads sent are already present.
2. It includes the `peer` in the callback. `replicated` does not tell you which peer was replicated.

This solves the issue I described in https://github.com/orbitdb/orbit-db/issues/509.

### DIFF
--- a/API.md
+++ b/API.md
@@ -60,6 +60,7 @@ Read the **[GETTING STARTED](https://github.com/orbitdb/orbit-db/blob/master/GUI
   * [`ready`](#ready)
   * [`write`](#write)
   * [`closed`](#closed)
+  * [`peer.exchanged`](#peerexchanged)
 
 <!-- tocstop -->
 
@@ -560,3 +561,10 @@ Emitted once the database has finished closing.
 ```javascript
 db.events.on('closed', (dbname) => ... )
 ```
+
+### `peer.exchanged`
+```javascript
+db.events.on('peer.exchanged', (peer, address, heads) => ... )
+```
+
+Emitted after heads have been exchanged with a peer for a specific database. This will be emitted after every exchange, even if no heads are received from the peer, or if all received heads are already present. (This is in contrast with the `replicated` event, which will only fire when new heads have been received.) Note that `heads` here contains heads *received* as part of the exchange, not heads sent.
--- a/src/OrbitDB.js
+++ b/src/OrbitDB.js
@@ -156,13 +156,14 @@ class OrbitDB {
   }
 
   // Callback for receiving a message from the network
-  async _onMessage (address, heads) {
+  async _onMessage (address, heads, peer) {
     const store = this.stores[address]
     try {
       logger.debug(`Received ${heads.length} heads for '${address}':\n`, JSON.stringify(heads.map(e => e.hash), null, 2))
       if (store && heads && heads.length > 0) {
         await store.sync(heads)
       }
+      store.events.emit('peer.exchanged', peer, address, heads)
     } catch (e) {
       logger.error(e)
     }
@@ -175,7 +176,7 @@ class OrbitDB {
     const getStore = address => this.stores[address]
     const getDirectConnection = peer => this._directConnections[peer]
     const onChannelCreated = channel => this._directConnections[channel._receiverID] = channel
-    const onMessage = (address, heads) => this._onMessage(address, heads)
+    const onMessage = (address, heads) => this._onMessage(address, heads, peer)
 
     const channel = await exchangeHeads(
       this._ipfs,
--- a/src/OrbitDB.js
+++ b/src/OrbitDB.js
@@ -160,10 +160,12 @@ class OrbitDB {
     const store = this.stores[address]
     try {
       logger.debug(`Received ${heads.length} heads for '${address}':\n`, JSON.stringify(heads.map(e => e.hash), null, 2))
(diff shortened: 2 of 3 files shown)


tyleryasaka:
Not sure what the test failures are about. Can't reproduce them locally. 🤷‍♂️

haadcode:
@tyleryasaka thanks for the PR! 👍 

The PR looks good overall, but I'd love to understand what the use case is for adding this? What are you trying to achieve from the user's perspective? Is the main use case to let the user know when peers have exchanged heads (in general? with a specific peer? just to know peers have connected and can find each other?)?

haadcode:
Another question is if the `peer.exchange` should fire after the *head* of the db was synced or if the *full database* was synced?

Re. the failing tests, it was a false positive due to bug in go-ipfs (https://github.com/libp2p/go-libp2p-pubsub/issues/93) so they sometimes fail, sometimes pass. if you see the same error in the future as in the one that previously failed, go to CI and press "Rebuild" and that usually does it (the success rate is about 50% :)).

tyleryasaka:
@haadcode The idea is to know when the heads have been exchanged with a specific peer and added locally, as well as any subsequent entries that are derived from those heads. Sometimes there will be no entries to add, (e.g. if they are already known).

Basically, I just want to know when I've been "caught up" with peer X (X being any peer I've explicitly connected to).

This can be especially useful in a setup where there is a primary node (or cluster of nodes) that are essentially acting as "servers", serving out the latest content to "clients". This is probably not the ideal setup for orbit-db (a bit too centralized), but at the moment it does seem to be a practical way to use orbit-db.

> Another question is if the peer.exchange should fire after the head of the db was synced or if the full database was synced?

Yeah. At first I was just firing the event after the heads were synced, but that was not so useful. Now, this fires after the entire database has synced. This is the point that it has been truly "caught up", and I can safely render the content to the user.

The reason that it's important to me to know this information has to do with usability. I would like users to see a loading spinner until content has been loaded from a trusted peer(s) that I feel have a high probability of having the latest content for a database (e.g. a centralized server). Without knowing whether the latest content has been loaded, the only option is to keep re-rendering the page as new data becomes available. This leads to a shaky user experience, where users could feel a bit disoriented seeing data suddenly appear on the page. A good user experience is one that is more definite - the user should feel that they can easily see when content is loading and when the loading is finished.

oed:
Is there anything blocking this PR? Would be great to have this feature!

aphelionz:
@tyleryasaka @oed So sorry to have dropped the ball on this. Can we get tests that cover this feature? It would help us to review it, help you to break it, and help users understand how it all works w/r/t syntax and semantics.

tyleryasaka:
@aphelionz No worries. I've been taking a break from coding for a while so it's been a while since I've looked at orbit-db. I just added some tests. Let me know what you think! I wasn't able to make the tests completely precise because there seemed to be some sort of race condition going on. I think it probably happens because the `peer.exchanged` event currently is a bit non-deterministic; it waits for the queue to empty before firing, which opens up a race condition. This is intentional as I have discussed previously. Making it deterministic would require a lot of other refactoring.

aphelionz:
I now see the clear case for this feature! I'd love to get this merged now.

aphelionz:
@tyleryasaka @oed  Out of curiousity, how are you working around this currently?

oed:
So right now we have an overlay pubsub room that does two things.
1. Client publishes the address of their DB, pinning-node listens and opens requested DB
2. When pinning-node opens DB it also sends a message in the pinning-room with the number of oplog entries it currently has. 

This makes it possible for the client to have a callback that is only called once the entire DB has been replicated from the pinning-node.

However both that approach and `peer.exchanged` requires the client to know what the address of the pinning-node is. Otherwise a malicious peer could send a message telling the client that there are `n+1` messages, when in reality there are only `n`. This could cause the client to wait indefinitely for more messages to the DB.

A (rather clunky) way around the issue of the client needing to know the address of the pinning-node could be the following (here we assume that there is only one user who writes to the DB, which is what happens in 3Box):
1. For each update to the db the user signs a message containing the address and the number of oplog entries.
2. This message is sent over the overlay pubsub room, all pinning-nodes that have this DB open also stores this message (or rather only the latest of these messages, if multiple updates)
3. When a client opens the DB the pinning-node sends the latest of these messages over the pubsub room.
4. The client can check the signature of this message and be certain how many messages to wait for until invoking the callback function.

So in conclusion `peer.exchanged` might not actually help my use case.

aphelionz:
Testing this locally now and the orbit-db tests seem consistent with what we're seeing in master. I'd be ok with merging this cc @haadcode and @shamb0t

aphelionz:
Superceded by #763 and finally slated for a release!!

PR #516, opened 2018-11-29
docs(README.md): Add note about calls above fold

RichardLitt:
In [at least this comment](https://github.com/orbitdb/orbit-db/issues/509#issuecomment-442710847), it was not clear that there were calls. This might help. What do you think?


haadcode:
LGTM, thanks @RichardLitt 👍

PR #525, opened 2018-12-14
Add project's status information and disclaimer to README

haadcode:
This PR will add a note on OrbitDB's status to the README. This is to improve users understanding of what to expect using OrbitDB.


RichardLitt:
100%. Good addition. Thanks, @haadcode.

PR #495, opened 2018-11-09
New Access Controller system

shamb0t:
This PR will make OrbitDB use the latest ipfs-log and access-control modules

Continuing from #449 

Changes:
- use ```createInstance``` instead of constructor
- move public-key information from ```orbitdb.key``` to ```orbitdb.identity```
- add access-controller modules
- update documentation

PR #531, opened 2018-12-17
Fix examples

haadcode:
This PR will update the examples to use the latest APIs. It'll remove the `lib` folder and copying to it when building examples and rather use the dist builds of each modules directly from `node_modules`.

From the commit message:
- Use new orbitdb API
- Remove usage of lib and use dist builds of each dependency

### DIFF
--- a/conf/webpack.config.js
+++ b/conf/webpack.config.js
@@ -8,7 +8,7 @@ module.exports = {
   output: {
     libraryTarget: 'var',
     library: 'OrbitDB',
-    filename: './dist/orbitdb.min.js'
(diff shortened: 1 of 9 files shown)


haadcode:
When testing the examples between two browsers (Chrome and Firefox), I'm seeing some weird behaviour where the clients just hang for seconds and nothing is rendered on the screen. Looking at the profiler, things are happening while the client "hangs". No idea what's going on but will need to try and see in more details what's going on.

haadcode:
Ran `make rebuild` and the weird behaviour reported earlier is gone. I'm gonna say this is ready to merge (to #495).

PR #532, opened 2018-12-17
Feat/update doc

shamb0t:
Update Docs to reflect AC and IP changes

### DIFF
--- a/GUIDE.md
+++ b/GUIDE.md
@@ -67,7 +67,7 @@ const ipfs = new IPFS(ipfsOptions)
 
 ipfs.on('ready', () => {
   // Create OrbitDB instance
-  const orbitdb = new OrbitDB(ipfs)
+  const orbitdb = await OrbitDB.createInstance(ipfs)
 })
 ```
 
@@ -87,7 +87,7 @@ Then, create a database instance (we'll use Key-Value database in this example):
 ```javascript
 const ipfs = new IPFS()
 ipfs.on('ready', async () => {
-  const orbitdb = new OrbitDB(ipfs)
+  const orbitdb = await OrbitDB.createInstance(ipfs)
   const db = await orbitdb.keyvalue('first-database')
 })
 ```
@@ -113,7 +113,7 @@ For example:
 ```javascript
 const ipfs = new IPFS()
 ipfs.on('ready', async () => {
-  const orbitdb = new OrbitDB(ipfs)
+  const orbitdb = await OrbitDB.createInstance(ipfs)
   const db = await orbitdb.keyvalue('first-database')
   console.log(db.address.toString())
   // /orbitdb/Qmd8TmZrWASypEp4Er9tgWP4kCNQnW4ncSnvjvyHQ3EVSU/first-database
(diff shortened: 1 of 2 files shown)


GUIDE.md
```diff
@@ -165,21 +179,22 @@ If you want to give access to other peers to write to a database, you need to ge
...
 
-***Note!*** *OrbitDB currently supports only write-access and the keys of the writers need to be known when creating a database. That is, the access rights can't be changed after a database has been created. In the future we'll support read access control and dynamic access control in a way that access rights can be added and removed to a database at any point in time without changing the database address. At the moment, if access rights need to be changed, the address of the database will change.*
+***Note!*** *OrbitDB currently supports only dynamically adding write-access. That is, write-access cannot be revoked once added. In the future we'll support read access control and dynamic access control in a way that access rights can be added and removed to a database at any point in time without changing the database address. At the moment, if access rights need to be removed, the address of the database will change.*
```
haadcode: I think we could rephrase this:

*"In the future we'll support read access control and dynamic access control in a way that access rights can be added and removed to a database at any point in time without changing the database address."* 

to 

*"In the future* ***OrbitDB will*** *support* ***access revocation and*** *read access control."*

What do you think?

GUIDE.md
```diff
@@ -165,21 +179,22 @@ If you want to give access to other peers to write to a database, you need to ge
...
 
-Access rights are setup by passing an `access` object that defines the access rights of the database when created. OrbitDB currently supports write-access. The access rights are specified as an array of public keys of the peers who can write to the database.
+Access rights are setup by passing an `accessController` object that defines the access-controller type and access rights of the database when created. OrbitDB currently supports write-access. The access rights are specified as an array of public keys of the peers who can write to the database.
```
haadcode: Can we add (to the end): "The public keys to give access to can be retrieved from `identity.publicKey` of each peer." or smth along those line?

GUIDE.md
```diff
@@ -223,29 +240,70 @@ This can be done by adding a `*` to the write access array:
...
   console.log(db.address.toString())
   // /orbitdb/QmRrauSxaAvNjpZcm2Cq6y9DcrH8wQQWGjtokF4tgCUxGP/first-database
 })
 ```
 
 Note how the access controller hash is different compared to the previous example!
 
+#### Granting access after database creation
+
+To give access to another peer after the database has been created, you must set the access-controller `type` to an `AccessController` which supports dynamically adding keys such as `OrbitDBAccessController`.
```
haadcode: We could add a small clarification in *"...which supports dynamically adding keys such..."* that *"...which supports dynamically adding* ***write access*** *such..."*

GUIDE.md
```diff
@@ -223,29 +240,70 @@ This can be done by adding a `*` to the write access array:
...
+```
+
+#### Custom Access Controller
+
+You can create a custom access controller by implementing the `AccessController` [interface](https://github.com/orbitdb/orbit-db-access-controllers/blob/master/src/access-controller-interface.js) and adding it to the AccessControllers object before passing it to OrbitDB.
+
+```javascript
+class OtherAccessController extends AccessController {
+    static get type () { return 'othertype'} // Return the type for this controller
+    async canAppend(entry, identityProvider) {} // return true if entry can be added, identityProvider allows you to verify entry.identity
```
haadcode: Can we expand on the actual function here? For example:
```javascript
async canAppend(entry, identityProvider) {
  if (entry.payload === "hello world" && entry.identity.id === identity.id)
    return true

  return false
}
```

to give the user a bit more information *how* they could use the canAppend?

GUIDE.md
```diff
@@ -223,29 +240,70 @@ This can be done by adding a `*` to the write access array:
...
 
+#### Granting access after database creation
+
+To give access to another peer after the database has been created, you must set the access-controller `type` to an `AccessController` which supports dynamically adding keys such as `OrbitDBAccessController`.
+
+```javaScript
+db = await orbitdb1.feed('AABB', {
+  accessController: {
+    type: 'orbitdb', //OrbitDBAccessController
+    write: [id1.publicKey]
```
haadcode: Can we rename `id1` here to `identity1`? I think that would avoid the possible confusion with `identity.id`.

GUIDE.md
```diff
@@ -223,29 +240,70 @@ This can be done by adding a `*` to the write access array:
...
+  }
+})
+
+await db.access.grant('write', id2.publicKey) // grant access to id2
```
haadcode: Can we rename `id2` here to `identity2`? I think that would avoid the possible confusion with `identity.id`.

GUIDE.md
```diff
@@ -223,29 +240,70 @@ This can be done by adding a `*` to the write access array:
...
+
+You can create a custom access controller by implementing the `AccessController` [interface](https://github.com/orbitdb/orbit-db-access-controllers/blob/master/src/access-controller-interface.js) and adding it to the AccessControllers object before passing it to OrbitDB.
+
+```javascript
+class OtherAccessController extends AccessController {
+    static get type () { return 'othertype'} // Return the type for this controller
+    async canAppend(entry, identityProvider) {} // return true if entry can be added, identityProvider allows you to verify entry.identity
+    async grant (access, identity) {} // Logic for granting access to identity
+}
+let AccessControllers = require('orbit-db-access-controllers')
```
haadcode: Tiny thing, but it'd make the example slightly more readable if we added an empty line here (as well as between 286 and 287) as to "group" logical things together (AC definition, adding it to orbit, using orbit).

haadcode:
Thanks @shamb0t for the updates! This is ready to be merged to #495 👍

PR #453, opened 2018-09-13
Fix non-base58 character error on windows

durac:
Replace backslash used in windows paths with slash to ensure correct conversion into parts.
With these changes orbit-db should support windows.

Related to #274

PR #558, opened 2019-02-06
Windows support remove the path reference

BartKnucle:
Path.join is plateform dependant, correct the hash differences between windows and poisx plateform

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   },
   "main": "src/OrbitDB.js",
   "dependencies": {
-    "ipfs-pubsub-1on1": "~0.0.4",
+    "ipfs-pubsub-1on1": "BartKnucle/ipfs-pubsub-1on1",
     "ipld-dag-pb": "0.14.11",
     "localstorage-down": "^0.6.7",
     "logplease": "^1.2.14",
--- a/src/OrbitDB.js
+++ b/src/OrbitDB.js
@@ -1,6 +1,5 @@
 'use strict'
 
-const path = require('path')
 const EventStore = require('orbit-db-eventstore')
 const FeedStore = require('orbit-db-feedstore')
 const KeyValueStore = require('orbit-db-kvstore')
@@ -36,7 +35,7 @@ class OrbitDB {
       : new Pubsub(this._ipfs, this.id)
     this.stores = {}
     this.directory = directory || './orbitdb'
-    this.keystore = options.keystore || Keystore.create(path.join(this.directory, this.id, '/keystore'))
+    this.keystore = options.keystore || Keystore.create(this.directory + '/' + this.id + '/' + '/keystore')
     this.cache = options.cache || Cache
     this.key = this.keystore.getKey(this.id) || this.keystore.createKey(this.id)
     this._directConnections = {}
@@ -235,7 +234,7 @@ class OrbitDB {
     const manifestHash = await createDBManifest(this._ipfs, name, type, accessControllerAddress, onlyHash)
 
     // Create the database address
-    return OrbitDBAddress.parse(path.join('/orbitdb', manifestHash, name))
+    return OrbitDBAddress.parse('/orbitdb' + '/' + manifestHash, name)
   }
 
   /* Create and Open databases */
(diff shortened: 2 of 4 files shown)


BartKnucle:
Dunno why tests fails... Not related to the pr to me.

RichardLitt:
Wouldn't this hardcode POSIX slashes, making it not work on Windows, anyway? Perhaps I am missing something.

BartKnucle:
Nope, I can confirm it's working on Windows now. The path.join add backslashes that break the path logic.

BartKnucle:
I think we can imagine a better code but I had to make it work quickly ;)

aphelionz:
IIRC Windows "supports" backslashes as well as forward slashes... or at least it's rather permissive when it comes to parsing them in paths

BartKnucle:
Yeah the problème is that if you construct the hashs on poisx and on windows, they will be differents if you don't take a standart. I don't know if I'm clear ;)

RichardLitt:
Cool! 

Hmm. Is this the standard approach for Node packages - getting rid of path.join? That seems like such a common package that I'm worried we're missing something.

BartKnucle:
Again, there maybe a clever way to handle that. But I just pointed out the difference between the path construction and the the hashes of the databases on poisx and windows system. If windows users want to be able to use orbit-db on windows and linux, the change must be done. Same applies to the pubsub1on1 package. A PR is on the way too. Feel free to write a better code, I'm not a very good dev :)

RichardLitt:
Yeah! I'm not trying to impugn your code's worth. I just don't know if there is another solution, so I am asking questions. :)

BartKnucle:
The simple test that can be done: execute the same database creation code on windows 10 and in the browsers. You'll see differences on the DB hashes. This PR just solve that :)

RichardLitt:
Ok! @aphelionz Want me to test this on my windows box? I forget if you have one.

aphelionz:
I don't have one, so yes please! @RichardLitt

RichardLitt:
Ok. So, tried today - and didn't get very far. I'm having an issue with npm installing go-ipfs-dep, and there's an issue with `internalBinding` in internal/util/inspect.js:31 not being defined. @BartKnucle Have you gotten past these errors? 

I would add the logs - but my computer currently can't access the internet for an unknown reason. I'm sorry. Not a windows guy.

BartKnucle:
Hi @RichardLitt , I never had these errors. I'm using js-ipfs instance. You can find my project code [here](https://github.com/BartKnucle/dext-starter/tree/master/lib/modules) (ipfs.js and orbitdb.js) They run on nodejs and the browser. You'll have also to apply [this commit](https://github.com/orbitdb/orbit-db/pull/453/commits/b8ce9366b74ea6dce71e846dc98ba6d94ee434e1)  to fix the base 58 error under windows.
For the db sync to work you'll also have to apply [https://github.com/ipfs-shipyard/ipfs-pubsub-1on1/pull/6](this)
Don't hesitate if you have further questions.

RichardLitt:
Thanks, @BartKnucle! This is important work - getting it working on windows would be an awesome step, and something we need to do. Thanks for bearing with me. :) Will update you when I get my internet working - it may be a local issue.

BartKnucle:
Sorry for the commit on the package.json. I needed to use my git repo version that work on windows from my package.json and it seems that any change I make on my repo since the PR is sent to the orgininal master branch. Anyway now I or Anyone can test the PR using my fork.
Maybe a "windows" branch should help? I'm just discovering GIT and I may made mistakes... Sorry.

RichardLitt:
Sorry for not getting to this sooner!

I think a windows branch might make more sense. You'll need to open a new PR from a new branch. It's a bit tricky - you'll need to make a new branch with your commits on it, push to your fork, then go back to master, pull from master over your commits, and then rebase master to your new windows branch, and then push again and make a PR from there. Let me know if you get totally lost with this. :)  We were all new at some point!

BartKnucle:
I think it's ok on: https://github.com/orbitdb/orbit-db/pull/595
Tell me if there is a mistake and thank you very much for the git lesson :)

RichardLitt:
Looks good to me! Closing this in favor of #595.

Issue #549, opened 2019-01-29
Issue with ipfs daemon code example


RichardLitt:
```js
const db = await orbitdb.log('hello')
```

I'm getting a linting error that you can't use await like this, outside of an async function. Should we remove it from the example?

BartKnucle:
Can you link where you get the code sample please?

RichardLitt:
@BartKnucle This is from the README, [here](https://github.com/orbitdb/orbit-db#module-with-ipfs-daemon).

aphelionz:
I think we need to either assume that people will understand that it needs to be wrapped in an `async` function, or wrap the examples in an async IIFE like so:

```javascript
(async () => {
  const db = await orbitdb.log('hello')
})()
```

RichardLitt:
Can we just add a note that it needs to be wrapped in an async? I think the IIFE might introduce more confusions than not. But this might just be _me_ being confused, so I'm not sure it's necessary to change anything.

aphelionz:
Yeah, a note should be fine

PR #561, opened 2019-02-08
docs: remove await outside of an async function in example

RichardLitt:
See #549.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,8 +139,9 @@ const OrbitDB = require('orbit-db')
 
 const ipfs = IpfsApi('localhost', '5001')
 const orbitdb = new OrbitDB(ipfs)
-const db = await orbitdb.log('hello')
-...
+orbitdb.log('hello').then(db => {
+  // Do something with your db.
+})
 ```
 
 ## API


README.md
```diff
@@ -139,7 +139,7 @@ const OrbitDB = require('orbit-db')
 
 const ipfs = IpfsApi('localhost', '5001')
 const orbitdb = new OrbitDB(ipfs)
-const db = await orbitdb.log('hello')
+const db = orbitdb.log('hello')
```
shamb0t: `orbitdb.log` is an async function so it will return a promise here. Maybe `orbitdb.log('hello').then(db => // do stuff with db )` ?

shamb0t:
Thanks @RichardLitt! You're right the issue is the three lines in the example should be called within an async function. But `.log` is async and will return a promise

RichardLitt:
Thanks, Shams. How does this look?

README.md
```diff
@@ -139,8 +139,9 @@ const OrbitDB = require('orbit-db')
 
 const ipfs = IpfsApi('localhost', '5001')
 const orbitdb = new OrbitDB(ipfs)
-const db = await orbitdb.log('hello')
-...
+const db = orbitdb.log('hello').then(db => {
```
shamb0t: The call will still return a promise so I think we should get rid of the `const db = `

shamb0t:
@RichardLitt one last comment! :D

README.md
RichardLitt: Removed!

shamb0t:
Thank you @RichardLitt ! :+1:

Issue #560, opened 2019-02-07
429 (Too Many Requests) browser example errors in console


huhsame:
I install orbit-db repository and run.
example works on browser but there are lots of errors in console.

**GET** https://node0.preload.ipfs.io/api/v0/refs?r=true&arg=Qmb3vcXiX6dBHjWojh817SoVUdBmxiycrPs23ySeMRHT5v **429 (Too Many Requests)**

This message is repeated with different hash. 
here are captures.

![image](https://user-images.githubusercontent.com/9151382/52402720-59553580-2b08-11e9-9f22-b14012b74545.png)
![image](https://user-images.githubusercontent.com/9151382/52402770-768a0400-2b08-11e9-9266-3f8ed0726afd.png)

package version 
"go-ipfs-dep": "0.4.18",
 "ipfs": "~0.33.0",

and node 10
mac

aphelionz:
@huhsame Try this.

When instantiating IPFS use: 

```javascript
this.ipfs = new Ipfs({
  preload: { enabled: false },
  ....
}
```

RichardLitt:
@aphelionz What does preload do?

huhsame:
https://github.com/ipfs/js-ipfs#optionspreload

haadcode:
This would be great candidate for FAQ

PR #563, opened 2019-02-14
Add note about preloading data

RichardLitt:
See https://github.com/orbitdb/orbit-db/issues/560.


FAQ.md
```diff
@@ -49,6 +49,19 @@ However, this will change in the future as js-ipfs gets GC and we want to make s
...
+This happens when there's only one node with data available, and the system isn't able to effectively get all of the data it needs from it. in order to get around this, instantiate IPFS with preload enabled:
+
+```js
+this.ipfs = new Ipfs({
+  preload: { enabled: false },
+  // ...
+}
+```
+
+This will preload data on the remote nodes, spreading out the demand across multiple nodes to avoid hitting one node too much.
```
aphelionz: This is correct but we're actual _disabling_ preload in this instance because the preload nodes are sometimes down.

RichardLitt: I should have actually read that. Edited and pushed change.

aphelionz:
Much better, thank you!

RichardLitt:
@haadcode What do you suggest as a way to merge FAQ questions faster? How about one accept is enough?

PR #571, opened 2019-02-27
chore: Add maintainers to package.json

RichardLitt:
Upon merging this, access rights should be given to both @shamb0t and @aphelionz to enable them to publish this package. This is in accordance with the OrbitDB [npm policy](https://github.com/orbitdb/welcome/blob/master/npm-policy.md). All maintianers must have 2FA enabled before given access.

- [ ] Enable 2FA (@haadcode)
- [x] Enable 2FA (@shamb0t)
- [x] Enable 2FA (@aphelionz)
- [ ] Give access to @aphelionz (@haadcode)
- [ ] Give access to @shamb0t (@haadcode)

I could also have publish rights, as needed. So could @vvp, I think.

I want to test this process with this repository, first. If anyone has any comments about how this is supposed to go, raise them here!


aphelionz:
I have 2FA enabled!

PR #576, opened 2019-03-04
docs: Add question on encryption

RichardLitt:
Taken from @aphelionz on Gitter.


aphelionz:
Looks great, thanks for doing this

PR #606, opened 2019-04-18
Tyleryasaka peer exchanged

aphelionz:
Creating this PR fo CircleCI tests


oed:
Any update on this? 🙂

oed:
Ping @aphelionz, is this still something you are going to merge?

aphelionz:
Currently on hold until we can get more input and eyes on the problem being addressed.

oed:
Ok, what do you need? This would be a great feature to have!

aphelionz:
@chmanie - you and I discussed this PR in Berlin. Can refresh my memory and point us all in the right direction to get this feature to where it needs to be to merge?

cc @oed

chmanie:
I think this PR is fine. The one it requires (namely https://github.com/orbitdb/orbit-db-store/pull/38) is not IMHO. The event listeners there need to be cleaned up, otherwise there will be memory leaks.

aphelionz:
I believe this is superceded by #763

PR #609, opened 2019-04-23
feat: Accept identity as an option when opening a DB

oed:
This code adds the ability to pass an identity when you open a DB. This allows you to have multiple DBs open with different identities at the same time.

### DIFF
--- a/src/OrbitDB.js
+++ b/src/OrbitDB.js
@@ -168,8 +168,9 @@ let databaseTypes = {
       cache: cache,
       onClose: this._onClose.bind(this),
     })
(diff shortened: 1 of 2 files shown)


shamb0t:
Thanks for the PR @oed! Lgtm but the tests are currently failing, could be due to a change in package-lock (I dont believe a change in package is necessary). Can you remove the package-lock and retry?

oed:
Thanks @shamb0t! Not sure why it changed. Tests where working locally btw. Let me update!

shamb0t:
Thanks @oed! Good to merge now :+1:

PR #616, opened 2019-05-01
docs: Add await comment

### LINKED ISSUE #549: Issue with ipfs daemon code example

RichardLitt:
Closes #549


haadcode:
Thanks @RichardLitt! 👍

PR #617, opened 2019-05-02
feat: Enable creation of DBs with old address format.

oed:
This change enables the creation of DBs with the old address format, by passing an additional option to the accessController. 

The main reason for this change is that in 3Box we are not saving the addresses of the users DBs. Rather, we generate the addresses by opening the DB with it's name and the writing key. Without this change we can only open a new db where the users data won't be available.

This PR depends on https://github.com/orbitdb/orbit-db-io/pull/2 and https://github.com/orbitdb/orbit-db-access-controllers/pull/29


shamb0t:
Thanks @oed! This is superseded by #625

PR #623, opened 2019-05-09
Persistent connection in the keystore

aphelionz:
## Overview
This PR implements the [`orbit-db-storage-adapter`](https://github.com/orbitdb/orbit-db-storage-adapter) paradigm which has two main benefits:

1. Greater performance (see benchmarks below)
2. Easier Interop to redis, mongo, sqlite at the cache and keystore level via the storage adapter

### Benchmarks

Before:
```
$ node benchmarks/benchmark-add.js
Starting IPFS daemon...
95 queries per second, 95 queries in 1 seconds (Oplog: 95)
87 queries per second, 182 queries in 2 seconds (Oplog: 182)
80 queries per second, 262 queries in 3 seconds (Oplog: 262)
88 queries per second, 350 queries in 4 seconds (Oplog: 350)
87 queries per second, 437 queries in 5 seconds (Oplog: 437)
85 queries per second, 522 queries in 6 seconds (Oplog: 522)
83 queries per second, 605 queries in 7 seconds (Oplog: 606)
88 queries per second, 693 queries in 8 seconds (Oplog: 693)
81 queries per second, 774 queries in 9 seconds (Oplog: 774)
--> Average of 86.1 q/s in the last 10 seconds
```

After:
```
$ node benchmarks/benchmark-add.js                             
Starting IPFS daemon...                                        
451 queries per second, 451 queries in 1 seconds (Oplog: 451)  
490 queries per second, 941 queries in 2 seconds (Oplog: 941)  
494 queries per second, 1435 queries in 3 seconds (Oplog: 1435)
513 queries per second, 1948 queries in 4 seconds (Oplog: 1948)
440 queries per second, 2388 queries in 5 seconds (Oplog: 2388)
415 queries per second, 2803 queries in 6 seconds (Oplog: 2803)
416 queries per second, 3219 queries in 7 seconds (Oplog: 3219)
375 queries per second, 3594 queries in 8 seconds (Oplog: 3594)
432 queries per second, 4026 queries in 9 seconds (Oplog: 4026)
--> Average of 445.3 q/s in the last 10 seconds  
```

## Depends On
This PR depends on the following PRs:

- [x] https://github.com/orbitdb/orbit-db-cache/pull/12
- [x] https://github.com/orbitdb/orbit-db-identity-provider/pull/29
- [x] https://github.com/orbitdb/orbit-db-keystore/pull/30
- [x] https://github.com/orbitdb/orbit-db-store/pull/55
- [x] https://github.com/orbitdb/localstorage-level-migration/pull/4
- [x]  https://github.com/orbitdb/ipfs-log/pull/251

## TODO
- [x] Replication still fails intermittently due to underlying js-ipfs bug (still function)
- [x] Deal with failing v0 tests
- [x] Deal with skipped tests
- [x] Update docs

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,23 +15,26 @@
   "dependencies": {
     "cids": "^0.7.1",
     "ipfs-pubsub-1on1": "~0.0.4",
+    "is-node": "^1.0.2",
     "localstorage-down": "^0.6.7",
     "logplease": "^1.2.14",
     "multihashes": "^0.4.12",
-    "orbit-db-access-controllers": "~0.2.0",
-    "orbit-db-cache": "~0.2.4",
+    "orbit-db-access-controllers": "~0.2.2",
+    "orbit-db-cache": "orbitdb/orbit-db-cache#master",
     "orbit-db-counterstore": "~1.5.0",
     "orbit-db-docstore": "~1.5.0",
     "orbit-db-eventstore": "~1.5.0",
     "orbit-db-feedstore": "~1.5.0",
-    "orbit-db-identity-provider": "~0.1.0",
-    "orbit-db-io": "~0.1.0",
-    "orbit-db-keystore": "^0.2.1",
+    "orbit-db-identity-provider": "github:orbitdb/orbit-db-identity-provider#master",
+    "orbit-db-io": "^0.1.1",
+    "orbit-db-keystore": "github:orbitdb/orbit-db-keystore#master",
     "orbit-db-kvstore": "~1.5.0",
     "orbit-db-pubsub": "~0.5.5",
-    "orbit-db-store": "~2.6.0"
+    "orbit-db-storage-adapter": "^0.4.2",
+    "orbit-db-store": "github:orbitdb/orbit-db-store#master"
   },
   "devDependencies": {
+    "babel-cli": "^6.26.0",
     "babel-core": "^6.26.0",
     "babel-loader": "^7.1.2",
     "babel-plugin-transform-runtime": "^6.23.0",
@@ -39,11 +42,11 @@
     "babel-preset-env": "^1.7.0",
     "datastore-level": "0.10.0",
     "fs-extra": "^7.0.1",
-    "go-ipfs-dep": "~0.4.20",
     "ipfs": "~0.36.4",
     "ipfs-repo": "~0.26.6",
     "ipfsd-ctl": "~0.42.3",
-    "localstorage-level-migration": "^0.0.1",
+    "go-ipfs-dep": "aphelionz/npm-go-ipfs-dep",
+    "localstorage-level-migration": "github:orbitdb/localstorage-level-migration#master",
     "markdown-toc": "^1.2.0",
     "mocha": "^5.2.0",
     "p-each-series": "^1.0.0",
@@ -55,6 +58,7 @@
     "remark-cli": "^5.0.0",
     "remark-validate-links": "^7.0.0",
     "rimraf": "^2.6.2",
+    "standard": "^12.0.1",
     "webpack": "^4.25.1",
     "webpack-cli": "^3.1.2"
   },
@@ -73,6 +77,14 @@
     "build:debug": "webpack --config conf/webpack.debug.config.js --sort-modules-by size && mkdir -p examples/browser/lib && cp dist/orbitdb.js examples/browser/lib/orbitdb.js && cp dist/orbitdb.js.map examples/browser/lib/orbitdb.js.map",
     "build:docs/toc": "markdown-toc --no-first1 -i README.md && markdown-toc --no-first1 -i API.md && markdown-toc --no-first1 -i GUIDE.md && markdown-toc --no-first1 -i CHANGELOG.md && markdown-toc --no-first1 -i FAQ.md ",
     "build:es5": "babel src --out-dir ./dist/es5/ --presets babel-preset-env --plugins babel-plugin-transform-runtime"
+  },
+  "standard": {
+    "env": "mocha",
+    "ignore": [
+      "test/**",
+      "examples/**",
+      "benchmarks/**"
+    ]
   },
   "localMaintainers": [
     "haad <haad@haja.io>",
--- a/src/OrbitDB.js
+++ b/src/OrbitDB.js
@@ -1,5 +1,6 @@
 'use strict'
 
+const fs = require('fs')
 const path = require('path')
 const EventStore = require('orbit-db-eventstore')
 const FeedStore = require('orbit-db-feedstore')
@@ -15,9 +16,11 @@ const OrbitDBAddress = require('./orbit-db-address')
 const createDBManifest = require('./db-manifest')
 const exchangeHeads = require('./exchange-heads')
 const { isDefined, io } = require('./utils')
+const Storage = require('orbit-db-storage-adapter')
+const leveldown = require('leveldown')
 
 const Logger = require('logplease')
-const logger = Logger.create("orbit-db")
+const logger = Logger.create('orbit-db')
 Logger.setLogLevel('ERROR')
 
 // Mapping for 'database type' -> Class
@@ -26,26 +29,25 @@ let databaseTypes = {
   'eventlog': EventStore,
   'feed': FeedStore,
   'docstore': DocumentStore,
-  'keyvalue': KeyValueStore,
+  'keyvalue': KeyValueStore
 }
 
-  class OrbitDB {
-  constructor(ipfs, identity, options = {}) {
-    if (!isDefined(ipfs))
-      throw new Error('IPFS is a required argument. See https://github.com/orbitdb/orbit-db/blob/master/API.md#createinstance')
+class OrbitDB {
+  constructor (ipfs, identity, options = {}) {
+    if (!isDefined(ipfs)) { throw new Error('IPFS is a required argument. See https://github.com/orbitdb/orbit-db/blob/master/API.md#createinstance') }
 
-    if (!isDefined(identity))
-      throw new Error('identity is a required argument. See https://github.com/orbitdb/orbit-db/blob/master/API.md#createinstance')
+    if (!isDefined(identity)) { throw new Error('identity is a required argument. See https://github.com/orbitdb/orbit-db/blob/master/API.md#createinstance') }
 
     this._ipfs = ipfs
     this.identity = identity
     this.id = options.peerId
     this._pubsub = options && options.broker
-      ? new options.broker(this._ipfs)
+      ? new options.broker(this._ipfs) // eslint-disable-line
       : new Pubsub(this._ipfs, this.id)
     this.directory = options.directory || './orbitdb'
     this.keystore = options.keystore
-    this.cache = options.cache || Cache
+    this.caches = { 'default': options.cache }
+    this.storage = options.storage
     this.stores = {}
     this._directConnections = {}
     // AccessControllers module can be passed in to enable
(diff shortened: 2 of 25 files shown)


src/OrbitDB.js
```diff
@@ -41,11 +56,12 @@ let databaseTypes = {
     this.identity = identity
     this.id = options.peerId
     this._pubsub = options && options.broker
-      ? new options.broker(this._ipfs)
+      ? new options.broker(this._ipfs) // eslint-disable-line
       : new Pubsub(this._ipfs, this.id)
     this.directory = options.directory || './orbitdb'
+    this.storage = options.storage
```
shamb0t: Do we need to pass in `storage` in addition to `keystore` and `cache`? Can this be handled in`createInstance` and only pass the latter?

src/OrbitDB.js
```diff
@@ -59,19 +75,25 @@ let databaseTypes = {
 
     const { id } = await ipfs.id()
     const directory = options.directory || './orbitdb'
-    const keystore = options.keystore || Keystore.create(path.join(directory, id, '/keystore'))
+    const storageOptions = Object.assign({}, options.storage, {
+      keystore: await storage.createStore(path.join(directory, id, '/keystore')),
```
shamb0t: Could this be made a function which is only called if a keystore isn't passed? Currently this store will be created regardless of whether a keystore is passed. Also if the passed keystore uses the same path we will get the `LOCK: held by other process` error. What do you think?

src/OrbitDB.js
```diff
@@ -239,7 +265,7 @@ let databaseTypes = {
     delete this.stores[address]
   }
 
-  async _determineAddress(name, type, options = {}) {
+  async _determineAddress(name, type, options = {}, onlyHash) {
```
shamb0t: Why remove `onlyHash` from the options? It needs to be passed to `createDBManifest`. I believe its an artifact that we have a `_determineAddress` and `determineAddress` I think we can get rid of one (I could be missing sth)

src/OrbitDB.js
```diff
@@ -369,22 +394,22 @@ let databaseTypes = {
 
   // Save the database locally
   async _addManifestToCache (directory, dbAddress) {
-    const cache = await this._loadCache(directory, dbAddress)
-    await cache.set(path.join(dbAddress.toString(), '_manifest'), dbAddress.root)
+    // const cache = await this._loadCache(directory, dbAddress)
```
shamb0t: `directory` can be removed from args

src/OrbitDB.js
aphelionz: Yes this is probably my least favorite part of this refactor. I think your idea is sound. Let me see what I can come up with!

src/OrbitDB.js
aphelionz: sorry, removed from the args of which function call? `_addManifestToCache`?

shamb0t: Oops sorry yea I meant from `_addManifestToCache`

src/OrbitDB.js
aphelionz: Have a look! I think it looks pretty good now.

aphelionz: How's that look?

src/OrbitDB.js
```diff
@@ -15,6 +16,20 @@ const OrbitDBAddress = require('./orbit-db-address')
...
+if (isNode) {
+  const leveldown = require('leveldown')
+  storage = Storage(leveldown)
+  storage.preCreate = async (directory, options) => {
+    fs.mkdirSync(directory, { recursive: true })
+  }
+} else {
+  const leveljs = require('level-js')
+  storage = Storage(leveljs)
+}
```
haadcode: Would it make sense to move the setup of the storage to `createInstance`? That way, we could add `storage` to the options in there and the user would be able to pass in what now is `leveldown` in the above code. I think that would get us directly to the point where this is customizable for the user (and we provide the default, as we do for identity and access controllers elsewhere in the code). What do you think?

src/OrbitDB.js
```diff
@@ -59,19 +74,32 @@ let databaseTypes = {
...
+      let cacheStorage = await storage.createStore(path.join(directory, id, '/cache'))
+      options.cache = new Cache(cacheStorage)
+    }
+
+    const finalOptions = Object.assign({}, options, {
       peerId: id ,
       directory: directory,
-      keystore: keystore
+      keystore: options.keystore,
+      cache: options.cache
```
haadcode: Are these two in fact needed? It seems to me we already assign new values to `options.cache` and `.keystore` above.

src/OrbitDB.js
```diff
@@ -59,19 +74,32 @@ let databaseTypes = {
...
+    }
+
+    if(!options.cache) {
+      let cacheStorage = await storage.createStore(path.join(directory, id, '/cache'))
+      options.cache = new Cache(cacheStorage)
+    }
+
+    const finalOptions = Object.assign({}, options, {
       peerId: id ,
       directory: directory,
```
haadcode: Iirc `directory` was removed elsewhere in the file, is this still needed?

src/OrbitDB.js
```diff
@@ -114,8 +142,12 @@ let databaseTypes = {
 
   async disconnect () {
     //close Keystore
-    if (this.keystore.close)
-      await this.keystore.close()
+    if (this.keystore._store.close)
+      await this.keystore._store.close()
```
haadcode: I think these (this and the one for cache below) would be nicer if they were just:
`await this.keystore.close()`

Not sure how to achieve that though...

src/OrbitDB.js
```diff
@@ -210,10 +240,10 @@ let databaseTypes = {
 
     const getStore = address => this.stores[address]
     const getDirectConnection = peer => this._directConnections[peer]
-    const onChannelCreated = channel => this._directConnections[channel._receiverID] = channel
+    const onChannelCreated = channel => this._directConnections[channel._receiverID] = channel // eslint-disable-line
```
haadcode: Nit-picking, but it'd be generally better to fix the code style rather than adding // eslint-disable.line. Here, iirc (I may not), the fixed style is:
```js
const onChannelCreated = channel => (this._directConnections[channel._receiverID] = channel)
```

src/orbit-db-address.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 const path = require('path')
-const multihash = require('multihashes')
+// const multihash = require('multihashes')
```
haadcode: Remove the line instead of commenting it out?

test/create-open.test.js
```diff
@@ -141,13 +141,6 @@ Object.keys(testAPIs).forEach(API => {
           assert.equal(manifest.accessController.indexOf('/ipfs'), 0)
         })
 
-        it('can pass local database directory as an option', async () => {
-          const dir = './orbitdb/tests/another-feed'
-          db = await orbitdb.create('third', 'feed', { directory: dir })
-          localDataPath = path.join(dir, db.address.root, db.address.path)
-          assert.equal(fs.existsSync(localDataPath), true)
-        })
-
```
haadcode: Why remove this test?

aphelionz: Since the store is persistent, we no longer pass the directory in at `orbitdb.open` time, instead passing it at `OrbitDB.createInstance` time.

src/OrbitDB.js
shamb0t: Couldnt `this._store.close()` be called in `keystore.close()`?

src/OrbitDB.js
```diff
@@ -295,8 +319,7 @@ let databaseTypes = {
   }
 
   async determineAddress(name, type, options = {}) {
-    const opts = Object.assign({}, { onlyHash: true }, options)
```
shamb0t: `onlyHash` should be added to the options

src/OrbitDB.js
shamb0t: Lgtm! :+1:

src/OrbitDB.js
aphelionz: That's correct - I'll re-implement that function inside `orbit-db-keystore` and `orbit-db-cache`. I was trying to avoid those modules controlling the storage at that level, but eh... 

Interacting directly vs interacting with "internal" variables at the `OrbitDb` level is six in one hand, half dozen in the other.

I'll implement `this.keystore.close`

src/OrbitDB.js
aphelionz: `directory` is needed at this point, and only this point. This is where we create and persistently connect to the store.

aphelionz:
@shamb0t @haadcode All tests are passing except _one_ replication test, and it's similar (though not the same) to the issues we're seeing in the v0 tests in `ipfs-log`. We're really close though.

I'm a bit spent today so I'm gonna take a rest but tomorrow I feel like we can together get this taken care of.

Note that I had to remove the usage of `MemStore` from `test/replicate.test.js` because it hadn't been updated to the base32 hashes. We can fix that up and restore it, or remove it completely.

src/OrbitDB.js
aphelionz: Good call. I simply removed the annotation in this branch since I'm doing all the linting in `fix/persisent-keystore-linting`. This one simply snuck in. I also removed the annotation from the other branch as well so that I remember to fix it :)

test/create-open.test.js
haadcode: Got it 👍

package.json
```diff
@@ -15,23 +15,26 @@
   "dependencies": {
     "cids": "^0.5.7",
     "ipfs-pubsub-1on1": "~0.0.4",
+    "is-node": "^1.0.2",
     "localstorage-down": "^0.6.7",
```
haadcode: Just noticed this: this should be a devDependency as its only used in tests.

src/OrbitDB.js
```diff
@@ -58,20 +61,39 @@ let databaseTypes = {
...
+      options.directory = './orbitdb'
+
+    if (!options.storage) {
+      const leveldown = isNode ? require('leveldown') : require('level-js')
+      options.storage = Storage(leveldown)
+      options.storage.preCreate = isNode ? async (directory, options) => {
+        fs.mkdirSync(directory, { recursive: true })
+      } : null;
+    }
+
```
haadcode: Thanks for changing this! 👍 

I now realize there's another, slightly more profound problem here with the browser build. As the check for whether we're in Node or browser happens dynamically (=runtime), webpack can't recognize it (when including OrbitDB as a module, the webpack config for building orbitdb does replace 'leveldown' with 'level-js') and will include both `leveldown` and `level-js` for the user. In [orbit-db-cache](https://github.com/orbitdb/orbit-db-cache), we separated the index/main files between browser and nodejs to solve that. I think we should find a solution for this, possibly similar to what cache does.

You can verify the inclusion by adding `--display-modules` to the webpack options in [build:examples](https://github.com/orbitdb/orbit-db/pull/623/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json&owned-by%5B%5D=#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R72) and [build:dist](https://github.com/orbitdb/orbit-db/pull/623/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json&owned-by%5B%5D=#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R73) and run them `npm run build:dist |grep level` (observe leveldown is not included) and `npm run npm run build:examples |grep level` (observe leveldown is included).

What do you think?

src/OrbitDB.js
```diff
@@ -58,20 +61,39 @@ let databaseTypes = {
...
-    })
-    const orbitdb = new OrbitDB(ipfs, identity, options)
-    return orbitdb
+    if(!options.directory)
+      options.directory = './orbitdb'
+
+    if (!options.storage) {
+      const leveldown = isNode ? require('leveldown') : require('level-js')
+      options.storage = Storage(leveldown)
+      options.storage.preCreate = isNode ? async (directory, options) => {
```
haadcode: Since [Storage](https://github.com/orbitdb/orbit-db-storage-adapter/blob/master/src/index.js#L28) takes in `options`, how about passing the `preCreate` function as an option instead of assigning it here? I feel that would be a bit cleaner and doesn't require overriding the function (and we can remove https://github.com/orbitdb/orbit-db-storage-adapter/blob/master/src/index.js#L65).

src/OrbitDB.js
haadcode: As a side note, it would make since to change https://github.com/orbitdb/orbit-db-storage-adapter/blob/master/src/index.js#L31 to use `await` instead of assuming preCreate is a Promise. That way it works even if the preCreate function is sync or not-a-Promise.

aphelionz: I believe this is all set, and the browser versions are building and examples are running locally for me :+1: :tada:

shamb0t: Browser examples lgtm locally too :+1:

test/counterdb.test.js
```diff
@@ -71,6 +71,7 @@ Object.keys(testAPIs).forEach(API => {
       let address
 
       it('creates and opens a database', async () => {
+        debugger;
```
haadcode: Can prolly remove this?

src/OrbitDB.js
```diff
@@ -296,7 +322,7 @@ let databaseTypes = {
 
   async determineAddress(name, type, options = {}) {
     const opts = Object.assign({}, { onlyHash: true }, options)
-    return this._determineAddress(name, type, opts)
+    return this._determineAddress(name, type, options, true)
```
shamb0t: This line should be unchanged (`opts` should be passed)

PR #624, opened 2019-05-15
Bumped NodeJS version

basraven:
- Bumped NodeJS version to 10 (container was building but not running any of the examples with the other version)
- Added SSL ignore to fix ssl issues when performing npm install

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:carbon
+FROM node:10
 
 ENV NPM_CONFIG_PREFIX=/home/node/.npm-global
 


basraven:
- ci/circle CI is failing due to broken mocha tests, not due to this change
- ci/circle CI **already uses the bumped version 10** of NodeJS, this Dockerfile update conforms to this.
- It aligns the Dockerfile with the Mocha tests NodeJS version (NVM10 is used in ci/circle).

shamb0t:
Thanks @basraven! The failure was due to a race condition in one of the tests, they are passing now :+1:  Can you elaborate on adding ssl ignore? What issues are you facing with npm install?

basraven:
> Thanks @basraven! The failure was due to a race condition in one of the tests, they are passing now 👍 Can you elaborate on adding ssl ignore? What issues are you facing with npm install?

The npm install will complain about self-signed certificates when running npm install, prefix like such will  ignore this (once). This will not happen when running the docker build for a second time due to caching in Docker (I expect), this is why it might have been missed. Happy to help!

```npm ERR! code SELF_SIGNED_CERT_IN_CHAIN
npm ERR! errno SELF_SIGNED_CERT_IN_CHAIN
npm ERR! request to https://registry.npmjs.org/lodash failed, reason: self signed certificate in certificate chain

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/node/.npm/_logs/2019-05-15T08_47_21_915Z-debug.log`
```

vvp:
Thanks for the PR @basraven ! :slightly_smiling_face:  Version bump is definitely ok but unfortunately we cannot disable the TLS certificate validation for security reasons. 

Also, I built this image myself without the NODE_TLS_REJECT_UNAUTHORIZED, and npm install was successful. Therefore it seems that the problem you're seeing may be caused by something else. Perhaps some HTTPS proxy doing certificate manipulation?

basraven:
> Thanks for the PR @basraven ! 🙂 Version bump is definitely ok but unfortunately we cannot disable the TLS certificate validation for security reasons.
> 
> Also, I built this image myself without the NODE_TLS_REJECT_UNAUTHORIZED, and npm install was successful. Therefore it seems that the problem you're seeing may be caused by something else. Perhaps some HTTPS proxy doing certificate manipulation?

Did you do the build with a purged Docker environment?
It only occurs with a purged Docker environment (on Windows 10).
I'm not behind an https proxy, so this is interesting.

For now: I isolated the NodeJS version bump and force pushed it to have a clean isolated version bump PR.
Let's ignore the SSL issue for now.

aphelionz:
I think this is OK to merge now, I see no issue with node:10 :+1:

shamb0t:
Happy to merge this now, thank you @basraven! I created an issue to track the certificate error you're getting https://github.com/orbitdb/orbit-db/issues/626

PR #625, opened 2019-05-15
pass format in dbManifest creation

shamb0t:
Allows `dag-pb` format to be used in creating manifest hashes

Uses `identiity.id` as default access-controller key

depends on https://github.com/orbitdb/orbit-db-io/pull/3 and https://github.com/orbitdb/orbit-db-access-controllers/pull/32


aphelionz:
Looks good! :+1:

PR #627, opened 2019-05-15
Examples folder default permission

basraven:
When running the container and building the examples (on Windows 10), the following error occurs:
```
Error: EACCES: permission denied, open '/home/node/examples/browser/browser-webpack-example/bundle.js'
```

Commands used to test this fix:
```
docker build -t orbit-db -f docker/Dockerfile . --no-cache
docker run -ti --rm orbit-db npm run build:examples
```

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -3,12 +3,14 @@ FROM node:10
 ENV NPM_CONFIG_PREFIX=/home/node/.npm-global
 
 WORKDIR /home/node
-USER node
 
 COPY package.json ./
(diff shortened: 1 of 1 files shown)


RichardLitt:
Sounds like a decent enough merge, to me. @shamb0t 	What do you think?

shamb0t:
Woops sorry to miss this...I'm not so familiar with Docker but the commands above appear to fix the error so happy to merge, thanks @basraven!

RichardLitt:
Thanks, @basraven !

PR #633, opened 2019-05-24
Set up CI with Azure Pipelines

shamb0t:


MatthewSteeples:
Will it also build on 2019 (just to keep dependencies up to date)?

RichardLitt:
@shamb0t 	Anything this needs to be merged?

shamb0t:
@RichardLitt this PR was auto-created by azure when we added automatic tests, they weren't passing so not to be merged yet. However, I believe that CI now provides windows environments so it will be more convenient to use that instead of Azure and this PR can be replaced and closed.

@MatthewSteeples thanks for the interest! You can ignore this PR, it is simply for adding azure windows test integration. Windows is still not fully supported but the closest thing right now is contents of #607.

PR #662, opened 2019-08-05
fix: assert value exists

alanshaw:
Fixes:

```
     TypeError [ERR_MISSING_ARGS]: The "actual" and "expected" arguments must be specified
      at Function.notEqual (assert.js:406:11)
      at Context.<anonymous> (test/create-open.test.js:137:18)
```


shamb0t:
Thank you @alanshaw!

PR #674, opened 2019-09-01
Preliminary cache migration code

aphelionz:
This PR adds `OrbitDb.attemptMigration`

This handles the automatic migration the cache from multiple LevelDB stores (pre-0.21) to a single LevelDB store with multiple keys. (0.22+)

- [X] Tests for "default cache" scenario
- [x] Tests for passing the `{ directory }` option into `OrbitDB.create`


src/OrbitDB.js
```diff
@@ -385,6 +387,40 @@ class OrbitDB {
     logger.debug(`Saved manifest to IPFS as '${dbAddress.root}'`)
   }
 
+  async _attemptMigration(options, dbAddress) {
+    let oldCache = this.caches[options.directory], oldStore
```
haadcode: I'm confused of the `,` after the assignment. How about declaring `oldStore` explicitly and putting on its own line?

haadcode:
Looks good to me, generally. One little stylistic comment. I'm still wondering though, if it would be possible to move the cache migration to its own file, and isolate it completely out of the main class?

aphelionz:
@haadcode Maybe we should start implementing a proper migrations folder. We could start with @shamb0t's 0.19-0.20, 0.20-0.21 migrations and include this as 0.21-0.22

aphelionz:
And then simply have some call to this.migrate() at places in the OrbitDB class

haadcode:
@aphelionz that'd make sense 👍

aphelionz:
@haadcode @shamb0t How's that look?

haadcode:
> @haadcode @shamb0t How's that look?

LGTM 👍

shamb0t:
LGTM too :+1:

src/migrations/0.21-0.22.js
```diff
@@ -0,0 +1,44 @@
...
+const fs = require('fs')
+
+const Cache = require('orbit-db-cache')
+
+const Logger = require('logplease')
+const logger = Logger.create('orbit-db')
+Logger.setLogLevel('ERROR')
+
+async function migrate (OrbitDB, options, dbAddress) {
+  let oldCache = OrbitDB.caches[options.directory]; let oldStore
```
shamb0t: Any reason not to put `let oldStore` on its own line?

aphelionz: ha, actually `standard --fix` changed that from a comma to a semicolon and left it on the same line

aphelionz: I'll move it down

shamb0t: Oh sneaky `standard`

PR #675, opened 2019-09-01
Ensure webpack compiles in downstream projects

aphelionz:
This PR removes `leveldown` as a direct dependency, instead relying on the implementation of `level` in `orbit-db-storage-adapter`, which in turn handles browser vs node dependency switching.

In addition, it utilizes a neat trick that uses eval to require the `fs` module, ensuring webpack compilation while still allowing node to use `fs`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12354,6 +12354,17 @@
         "logplease": "~1.2.15",
         "mkdirp": "^0.5.1",
         "orbit-db-storage-adapter": "^0.4.1"
+      },
+      "dependencies": {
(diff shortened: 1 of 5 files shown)


shamb0t:
Not sure I understand the need to use the `eval` trick. Shouldn't this line https://github.com/orbitdb/orbit-db/blob/9b58f429cdfb4ae03d1278cdf08ef3fa776cfb83/conf/webpack.config.js#L16 allow webpack to compile?

shamb0t:
Ah got it for other projects to compile using webpack

PR #676, opened 2019-09-03
Fix/tests

aphelionz:


haadcode:
@aphelionz give the original date of the issue, is this still valid? If not, let's close it as obsolete.

aphelionz:
Agreed

PR #677, opened 2019-09-03
Using identity's keystore as default first

aphelionz:
package-lock


src/fs-shim.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
 const fs = (typeof window === 'object' || typeof self === 'object') ? null
   : eval('require("fs")')
```
aphelionz: `eval` considered harmful, except that this eval has no user input associated with it

phillmac: Just for the sake of my curiosity, what does the eval even do here?

aphelionz: in the event of it not being in the browser (no `window` object), or a web worker (no `self` object) it simply calls the result of evaluating `require("fs")`, which is... `require('fs')`

It's a "clever" runtime check for fs and it allows downstream projects to not have to deal with `fs` in their own webpack configs

phillmac: Makes sense. I didn't catch on how it prevents an error in the browser with the undefined `require` function

src/OrbitDB.js
```diff
@@ -45,16 +45,23 @@ class OrbitDB {
...
     this.directory = options.directory || './orbitdb'
-    this.keystore = options.keystore
-    this.caches = { 'default': options.cache }
     this.storage = options.storage
-    this.stores = {}
     this._directConnections = {}
+
+    this.caches = { 'default': options.cache }
+    // flag so that the store class doesn't close it
+    if(this.caches['default']) this.caches['default'].isDefault = true
```
haadcode: Is `isDefault` used anywhere?

src/OrbitDB.js
```diff
@@ -75,6 +82,10 @@ class OrbitDB {
       options.storage = Storage(null, storageOptions)
     }
 
+    if (options.identity && options.identity.provider._keystore) {
```
haadcode: We should really expose `_keystore` as `keystore` in the identity provider. The member variables prefixed with `_` are not meant to be called directly (as in they are a notion of "private variables", as far as that goes in JS).

src/OrbitDB.js
```diff
@@ -137,20 +148,20 @@ class OrbitDB {
...
+    }
+    this.keystores = []
 
     // Close all open databases
     const databases = Object.values(this.stores)
     for (let db of databases) {
       await db.close()
       delete this.stores[db.address.toString()]
     }
+    await this.caches['default'].close()
```
haadcode: Why only close the default cache and not others?

src/OrbitDB.js
```diff
@@ -295,10 +305,16 @@ class OrbitDB {
     // Create the database address
     const dbAddress = await this._determineAddress(name, type, options)
 
-    options.cache = this.caches[options.directory || 'default']
     if (!options.cache) {
-      const cacheStorage = await this.storage.createStore(options.directory)
-      this.caches[options.directory] = options.cache = new Cache(cacheStorage)
+      if (options.directory) {
+        const cacheStorage = await this.storage.createStore(options.directory)
+        this.caches[options.directory] = options.cache = new Cache(cacheStorage)
```
haadcode: Can this be broken into multiple lines instead of chaining? I think that would make it more readable and less error prone.

src/OrbitDB.js
```diff
@@ -295,10 +305,16 @@ class OrbitDB {
...
+        if (options.cache._store.db.status !== 'open') {
+          await options.cache._store.open()
```
haadcode: Can you help me understand why we need to check the store status directly here? It feels like this should be handled on another layer (in the store/storage modules?)...

src/OrbitDB.js
```diff
@@ -353,7 +369,10 @@ class OrbitDB {
     // Parse the database address
     const dbAddress = OrbitDBAddress.parse(address)
 
-    if (!options.cache) options.cache = this.caches['default']
+    options.cache = this.caches[options.directory] || this.caches['default']
+    if (options.cache._store.db.status !== 'open') {
+      await options.cache._store.open()
```
haadcode: Same question as in https://github.com/orbitdb/orbit-db/pull/677/files?file-filters%5B%5D=.js#r321580114.

src/OrbitDB.js
aphelionz: Yes, in `orbit-db-store` here: https://github.com/orbitdb/orbit-db-store/blob/master/src/Store.js#L193

Basically this prevents `db.drop` and `db.close` from inadvertently closing the "master" or "default" store managed by OrbitDB. This is in line with the overall refactor of not creating a separate cache store for every DB _unless_ `options.directory` is passed into `OrbitDB.create` or `OrbitDB.open`, which is another performance optimization

src/OrbitDB.js
aphelionz: Actually this is already implemented as a getter function in `orbit-db-identity-provider` so I'll just use that, np.

src/OrbitDB.js
aphelionz: The idea here is that the for loop above will call `db.close()` on each store, which will close any "custom" caches created via the `options.directory`. However, since writing this code I implemented a `this.cache` getter in `OrbitDB`, so I'll use that.

src/OrbitDB.js
aphelionz: :+1:

src/OrbitDB.js
aphelionz: This is a safeguard against trying to access a store too soon, before it is open. I put the check here to prevent an API breaking change.

Since constructors are not async we would have to migreate to an `async Cache.create` function and have people update their code. This way we can avoid that.

src/OrbitDB.js
aphelionz: Same answer :)

aphelionz:
@haadcode I believe all the tests should pass now

aphelionz:
Ok, pushed some more requested changed.

@haadcode @shamb0t Please have a look. I'm too deep in the weeds to think about it clearly / objectively and some of the requested changes require changes to our semantics around `drop` and `close`

You can see examples of this in the tests and also in this `orbit-db-cache` PR: https://github.com/orbitdb/orbit-db-cache/pull/14

src/OrbitDB.js
```diff
@@ -260,6 +274,23 @@ class OrbitDB {
...
+    // Cache cleanup
+    // First remove the handler by db address
+    // Then close the cache if there are no more handlers
+    for(let directory of Object.keys(this.caches)) {
+      const cache = this.caches[directory] || this.cache
+      const store = this.stores[address]
+
+      if (store && store.options.directory == directory) {
+        // Closing a custom cache?
+        if (cache.handlers.has(address)) cache.handlers.delete(address)
```
shamb0t: Can this be handled via some interface in `orbit-db-cache`? The `handlers` property shouldn't be directly manipulated from orbitdb imo

src/OrbitDB.js
```diff
@@ -260,6 +274,23 @@ class OrbitDB {
...
     }
 
+    // Cache cleanup
+    // First remove the handler by db address
+    // Then close the cache if there are no more handlers
+    for(let directory of Object.keys(this.caches)) {
+      const cache = this.caches[directory] || this.cache
+      const store = this.stores[address]
+
+      if (store && store.options.directory == directory) {
```
shamb0t: Maybe a typo but think we should stick with triple === throughout the codebase, any reason not to?

test/create-open.test.js
```diff
@@ -340,5 +337,49 @@ Object.keys(testAPIs).forEach(API => {
...
+        const db3 = await orbitdb.open('xyz3', { create: true, type: 'feed', directory  })
+        const db4 = await orbitdb.open('xyz4', { create: true, type: 'feed', directory: directory2  })
+        const db5 = await orbitdb.open('xyz5', { create: true, type: 'feed', })
+
+        await db1.close()
+        await db2.close()
+        await db4.close()
+
+        assert.strictEqual(orbitdb.cache._store._db.status, 'open')
+        assert.strictEqual(db2._cache._store._db.status, 'open')
```
shamb0t: Shouldn't this be 'closed'?

shamb0t:
@aphelionz can you provide tests which fail that this PR fixes? It's still not clear to me that this is necessary. Are these essentially keystore and cache managers? Is it inconsistent to expect users to close cache/keystore themselves if they are non-default ones passed in?

Also can you elaborate on 
> some of the requested changes require changes to our semantics around `drop` and `close`

What are the requested changes and what are the changes to the semantics around `drop` and `close`?

src/OrbitDB.js
shamb0t: Perhaps we could call sth like `this.cache.rmHandler(address)`, the check can be done in `rmHandler` and the cache can close itself if its handlers size is 0. What do you think?

src/OrbitDB.js
```diff
@@ -337,6 +375,19 @@ class OrbitDB {
...
+        options.cache = this.caches[options.directory]
+      } else {
+        this.caches[options.directory] = await this._createCache(options.directory)
+        options.cache = this.caches[options.directory]
+      }
+    } else {
+      options.cache = this.cache
+    }
+    // "Wake up" the caches if they need it
+    if (options.cache) await options.cache.open()
```
haadcode: Could this part be pulled out to its own function since we use exactly the same code above too?

test/create-open.test.js
haadcode: I believe this is correct as it is. `db2` and `db3` use the same cache (`directory`), so if `db3` is not closed yet, closing `db2` should keep the cache open for `db3` to keep using it. Makes sense?

test/replication-status.test.js
```diff
@@ -89,8 +89,9 @@ Object.keys(testAPIs).forEach(API => {
     })
 
     it('has correct replication info after loading from snapshot', async () => {
+      await db._cache._store.open();
       await db.saveSnapshot()
-      await db.close()
+      // await db.close()
```
haadcode: I'm unsure if we can remove this. The test here is supposed to "load the db form a snapshot" (as in "fresh"), so having everything closed as a starting point is probably better (in order to make sure it all works). We could possible move the `saveSnapshot` call to a `before(() => ...)` to make that clear?

test/replication-status.test.js
```diff
@@ -58,7 +58,7 @@ Object.keys(testAPIs).forEach(API => {
 
     it('has correct replication info after load', async () => {
       await db.add('hello')
-      await db.close()
+      // await db.close()
```
haadcode: Same comment as for https://github.com/orbitdb/orbit-db/pull/677/files#r325600547

src/OrbitDB.js
```diff
@@ -137,13 +139,8 @@ class OrbitDB {
...
   async disconnect () {
-    // close Keystore
-    await this.keystore.close()
-
-    // close Cache
-    await Promise.all(Object.values(this.caches).map((cache) => {
-      return cache.close()
-    }))
+    // close keystore
+    await this.keystore.close
```
shamb0t: Missing brackets in call to `close()`?

aphelionz: :+1:

PR #681, opened 2019-09-10
adds optional meta field to manifest

tabcat:
This PR would give users the ability to add additional data to the db manifest. This could be useful when wanting to package information with the db that should not change during the lifetime of the db. An example could be adding a public ECDH key that the owner of the db controls.

The meta field would only exist in the manifest if `options.meta !== undefined` is true when creating the database. This keeps db addresses the same as before when creating a db without a meta field. When creating a db with a meta field, make options.meta equal to a JSON serialize-able value so it can safely traverse to/from ipfs. This value with be readable in the db instance property db.options.meta (if the manifest does not contain a meta field db.options.meta is undefined).

The tests added will check when creating a db that 1) the meta field is only added to the manifest if options.meta is defined 2) if there is no meta field in the manifest db.options.meta is undefined 3) if there is a meta field when options.meta is defined and options.meta is equal to db.options.meta

### DIFF
--- a/src/db-manifest.js
+++ b/src/db-manifest.js
@@ -6,7 +6,9 @@ const createDBManifest = async (ipfs, name, type, accessControllerAddress, optio
   const manifest = {
     name: name,
     type: type,
-    accessController: path.join('/ipfs', accessControllerAddress)
+    accessController: path.join('/ipfs', accessControllerAddress),
+    // meta field is only added to manifest if options.meta is defined
+    ...(options.meta !== undefined ? { meta: options.meta } : {})
   }
 
   return io.write(ipfs, options.format || 'dag-cbor', manifest, options)
--- a/src/db-manifest.js
+++ b/src/db-manifest.js
@@ -3,13 +3,14 @@ const io = require('orbit-db-io')
 
 // Creates a DB manifest file and saves it in IPFS
 const createDBManifest = async (ipfs, name, type, accessControllerAddress, options) => {
-  const manifest = {
+  const manifest = Object.assign({
     name: name,
     type: type,
-    accessController: path.join('/ipfs', accessControllerAddress),
-    // meta field is only added to manifest if options.meta is defined
-    ...(options.meta !== undefined ? { meta: options.meta } : {})
-  }
+    accessController: path.join('/ipfs', accessControllerAddress)
+  },
+  // meta field is only added to manifest if options.meta is defined
+  options.meta !== undefined ? { meta: options.meta } : {}
+  )
 
   return io.write(ipfs, options.format || 'dag-cbor', manifest, options)
 }
--- a/test/create-open.test.js
+++ b/test/create-open.test.js
@@ -225,7 +225,7 @@ Object.keys(testAPIs).forEach(API => {
           it('creates a manifest with no meta field', async () => {
             db = await orbitdb.create('no-meta', 'feed')
             const manifest = await io.read(ipfs, db.address.root)
-            assert.strictEqual(db.options.meta, undefined)
+            assert.strictEqual(manifest.meta, undefined)
             assert.deepStrictEqual(Object.keys(manifest).filter(k => k === 'meta'), [])
           })
 
@@ -233,7 +233,7 @@ Object.keys(testAPIs).forEach(API => {
             const meta = { test: 123 }
             db = await orbitdb.create('meta', 'feed', { meta })
             const manifest = await io.read(ipfs, db.address.root)
-            assert.deepStrictEqual(db.options.meta, meta)
+            assert.deepStrictEqual(manifest.meta, meta)
             assert.deepStrictEqual(Object.keys(manifest).filter(k => k === 'meta'), ['meta'])
           })
         })
--- a/test/create-open.test.js
+++ b/test/create-open.test.js
@@ -207,6 +207,36 @@ Object.keys(testAPIs).forEach(API => {
             assert.deepEqual(db.access.write, [orbitdb.identity.id])
           })
         })
+        describe('Meta', function() {
+          before(async () => {
+            if (db) {
+              await db.close()
+              await db.drop()
+            }
+          })
+
+          afterEach(async () => {
+            if (db) {
+              await db.close()
+              await db.drop()
+            }
+          })
+
+          it('creates a manifest with no meta field', async () => {
+            db = await orbitdb.create('no-meta', 'feed')
+            const manifest = await io.read(ipfs, db.address.root)
+            assert.strictEqual(db.options.meta, undefined)
+            assert.deepStrictEqual(Object.keys(manifest).filter(k => k === 'meta'), [])
+          })
+
+          it('creates a manifest with a meta field', async () => {
+            const meta = { test: 123 }
+            db = await orbitdb.create('meta', 'feed', { meta })
+            const manifest = await io.read(ipfs, db.address.root)
+            assert.deepStrictEqual(db.options.meta, meta)
+            assert.deepStrictEqual(Object.keys(manifest).filter(k => k === 'meta'), ['meta'])
+          })
+        })
       })
     })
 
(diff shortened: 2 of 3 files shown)


src/db-manifest.js
```diff
@@ -3,11 +3,14 @@ const io = require('orbit-db-io')
...
 const createDBManifest = async (ipfs, name, type, accessControllerAddress, options) => {
-  const manifest = {
+  const manifest = Object.assign({
     name: name,
     type: type,
     accessController: path.join('/ipfs', accessControllerAddress)
-  }
+  },
+  // meta field is only added to manifest if options.meta is defined
+  options.meta !== undefined ? { meta: options.meta } : {}
```
aphelionz: What does it write if you simply set options.meta to null? Does it not just omit the field?

phillmac: Yeah I've encountered this omit behaviour a couple of times with the http-api, and I consider it to be undesirable in most cases because I don't wish to check if the intermediate object exists when using child properties. e.g.: 
`const abc = options.meta.something`  turn into
`const abc = options.meta && options.meta.something`
having a guarantee that an intermediate object exists even if it's empty is the better choice IMHO
I realise however that I have no say in the matter because it's not even my PR :P 
I guess the flip-side is trying to keep the manifest as lightweight as possible.
It's a trade-off like all things I guess.

tabcat: @aphelionz im assuming you mean to have the meta property set to options.meta in the manifest object and options.meta set to null if undefined beforehand? This defines the meta field as null in the object and is added to ipfs which changes the db address. The goal of this part of the code is to only add the meta field to the manifest if defined so creating dbs with the same config as before gives the same address.

phillmac: `options.meta = Object.assign({}, options.meta)`

tabcat: @phillmac thats fine but if options.meta is not an object there may be a problem. also how are you adding that to the manifest object?

phillmac: My bad I didn't quite understand what you were doing there

tabcat: I think it would be best if meta field is only added to the manifest if explicitly defined in options, primarily because of db addresses. if there is no meta field defined its defined as undefined in db.options.meta anyway

aphelionz: I agree, your way is backwards compatible. Just checking!

haadcode:
Thank you @tabcat for the PR! 🙏 This would indeed be a good addition and has been requested in the past by some users.

To understand the approach a little better, can you elaborate on the rationale for separating `defaults` and `meta` (as per https://github.com/tabcat/orbit-db/commit/da3ec816159a9c179ad162234063bed110041cd7#diff-f1c4721cce5f33ff9a226802788fe12aR384)? Are those not the same? Could they be the same, as in just have the `meta` field?

tabcat:
@haadcode They could indeed yes. I designed the meta field to be totally controlled by the user but it could encapsulate defaults and some field for the user by changing the code a bit. Options.defaults and options.meta could stay the same but be set up in the manifest differently (eg. meta: { defaults, meta }). both should still probably both be added optionally.

aphelionz:
I can see the distinction also being at the "OrbitDB layer" vs the "application layer". If you're just issuing a naked `orbitdb.open` call (like on a pinning service or something) you don't need to care about what's in `meta`. People are gonna do whatever they want anyway, I imagine, but it's still a worthy consideraton imo

tabcat:
Yeah I think I like separating them more but not sure there is really a difference anywhere but the manifest.

RichardLitt:
I like the idea. Stupid question: What's to stop me from adding a meta field to the data, without adding all of these checks? Can't I just add my own fields to any JSON object without having an issue?

tabcat:
if by checks you mean checking if meta is defined in options *before* adding the meta field to the manifest, the goal of the check is to enable backwards compatible and adding a new field everywhere even if its not used would break that property. It would work just fine but anyone using orbitdb without this feature would have different addresses with the same config because they would not have the meta field in their manifest.

tabcat:
if orbitdb ever needs to add something that breaks the manifest address anyway these checks can be removed and the meta field could be added by default with it but i didnt see a reason to make a breaking change here.

src/db-manifest.js
shamb0t: I think readability could be improved if this was compacted a little to `options.meta ? { meta: options.meta } : {}`

test/create-open.test.js
```diff
@@ -207,6 +207,36 @@ Object.keys(testAPIs).forEach(API => {
...
+              await db.close()
+              await db.drop()
+            }
+          })
+
+          it('creates a manifest with no meta field', async () => {
+            db = await orbitdb.create('no-meta', 'feed')
+            const manifest = await io.read(ipfs, db.address.root)
+            assert.strictEqual(db.options.meta, undefined)
+            assert.deepStrictEqual(Object.keys(manifest).filter(k => k === 'meta'), [])
```
shamb0t: Could we use `assert.strictEqual(manifest.meta, undefined)` here to match the convention in the rest of the codebase? What do you think?

test/create-open.test.js
```diff
@@ -207,6 +207,36 @@ Object.keys(testAPIs).forEach(API => {
...
+          })
+
+          it('creates a manifest with a meta field', async () => {
+            const meta = { test: 123 }
+            db = await orbitdb.create('meta', 'feed', { meta })
+            const manifest = await io.read(ipfs, db.address.root)
+            assert.deepStrictEqual(db.options.meta, meta)
+            assert.deepStrictEqual(Object.keys(manifest).filter(k => k === 'meta'), ['meta'])
```
shamb0t: For similar reasons perhaps here we could use `assert.notEqual(manifest.meta, undefined)`?

shamb0t:
Agrees this is a great addition, thank you @tabcat! Just commented a few stylistic suggestions

src/db-manifest.js
tabcat: my thought for using  `options.meta !== undefined` here was users being able to store any value they wanted including falsey values but thats not a big deal, ik i will almost only be using it to store objects.

test/create-open.test.js
tabcat: yes that's a good idea

test/create-open.test.js
tabcat: i guess either one works although using not notEqual here the next line, line 237, seems close to redundant. line 237's type of check is really only needed in the no-meta test, if you dont see a need for using deepStrictEqual here you could get rid of one of those lines.

test/create-open.test.js
tabcat: i guess the like 228 and 229 are also redundant?

aphelionz:
Happy to merge this once the feedback has been addressed

test/create-open.test.js
shamb0t: Hmm could be but I don't see them as redundant since you're checking whether `meta` is set in the db options, while 229 checks that `meta` has been written to ipfs.

src/db-manifest.js
shamb0t: Fair point! I think in that case what you have here makes sense

test/create-open.test.js
tabcat: made a commit to try and solve the code convention issue. it makes lines 228 and 229 sort of redundant so we could remove line 228, but personally it doesnt seem like a big deal either way and also keeps both tests more symmetrical. lmk what you think

test/create-open.test.js
tabcat: made a commit below that changes db.options.meta to manifest.meta in both tests. it seemed like it would solve the code convention issue here.

aphelionz:
@shamb0t satisfied with the code changes above?

shamb0t:
Thanks @tabcat @aphelionz  lgtm :+1:

PR #683, opened 2019-09-16
fix: exchanges heads with peer, once the heads are available

zachferland:
Fixes a bug where a peer connects before the db is 'ready' (before it has store._oplog.heads available), and  it doesn't exchanges any heads with the peer as a result. 

This listens for ready event on given store, and will exchange heads again if the store was not ready on original call. 

Seemed like simplest implementation of fix, but maybe better fix exist that manages this state and only sends once. Either way overhead seemed low, to just send redundantly in these cases.

### DIFF
--- a/src/exchange-heads.js
+++ b/src/exchange-heads.js
@@ -33,11 +33,19 @@ const exchangeHeads = async (ipfs, address, peer, getStore, getDirectConnection,
   logger.debug(`Connected to ${peer}`)
 
   // Send the heads if we have any
-  const heads = getHeadsForDatabase(getStore(address))
+  const store = getStore(address)
+  const heads = getHeadsForDatabase(store)
+
   logger.debug(`Send latest heads of '${address}':\n`, JSON.stringify(heads.map(e => e.hash), null, 2))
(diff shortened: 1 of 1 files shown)


aphelionz:
Thanks @zachferland. This would definitely be good to patch if it fixes the stated problem. 

However, the concurrent syncing tests fail - are they passing locally on master for you without those changes? Also, might you be able to add a failing test that this patch fixes, so we can avoid regressions?

zachferland:
@aphelionz ran the tests locally again, on both this branch and master. On master, browser-multiple-tabs failed on first run, but then pass every time after, 'npm run test:all' all passed. Then on this branch locally 'npm run test:all', all tests passed as well (multiple runs)

Glad to add a test here to catch regression, havent looked at test code base here in a while, any good spots to add a test like this?

aphelionz:
Ok @zachferland that test looks like our problem, not yours.

I think the appropriate place would be under `test/replicate-automatically.js`

aphelionz:
Correct me if I'm wrong but I believe this can be closed in favor of #688

PR #684, opened 2019-09-17
Write Access with id ?

spartanz51:


phillmac:
The id is not the publuc key. The public key is required for access control :cat:

tabcat:
weird, ive always added the id to the accessController.write array

tabcat:
these changes seem correct

phillmac:
It used to be that the id was the public key, but recently it's become possible to define the public key and id separately, so id !== public key

phillmac:
See  https://github.com/orbitdb/orbit-db-identity-provider/blob/e473da659b1ef657a77bf2314854dfb3838149ff/src/identity.js#L5

aphelionz:
Thanks for the PR @spartanz51 .

These changes are indeed correct. You can see the tests here: https://github.com/orbitdb/orbit-db/search?q=%22.identity.id%22&unscoped_q=%22.identity.id%22

There was another discussion about this recently as well, in this PR. You can follow more of the discussion there. https://github.com/orbitdb/orbit-db-access-controllers/pull/31

GUIDE.md
```diff
@@ -203,7 +203,7 @@ ipfs.on('ready', async () => {
 })
 ```
 
-To give write access to another peer, you'll need to get their public key with some means. They'll need to give you the output of their OrbitDB instance's key: `orbitdb.identity.publicKey`.
+To give write access to another peer, you'll need to get their public key with some means. They'll need to give you the output of their OrbitDB instance's key: `orbitdb.identity.id`.
```
tabcat: should probably have changed 'OrbitDB instance's key' to 'OrbitDB instance's id'

aphelionz:
Fixed it in post ;)

PR #685, opened 2019-09-18
Upgrade to js-IPFS v0.38.x-rc

achingbrain:
There's a new release of js-IPFS around the corner, one of the things it does is turn on pubsub by default so needs a little code tweaking to work.

This PR contains the changes required to get orbit-db working with the latest release candidate.


test/utils/test-apis.js
```diff
@@ -8,14 +9,16 @@ const IPFS = require('ipfs')
 // https://github.com/ipfs/js-ipfsd-ctl#ipfsfactory---const-f--ipfsfactorycreateoptions
 let jsIpfs = {
   'js-ipfs': {
-    type: 'proc', 
+    type: 'proc',
     exec: IPFS,
+    IpfsClient: IPFSHTTPClient,
```
aphelionz: Thanks for updating all of this!

Question here - probably something I don't fully understand but it seems counterintuitive to use the HTTP Client for `js-ipfs`? Is there something significant that changed under the hood?

aphelionz:
@achingbrain Also it looks like while tests are passing, there's a number of socket resets and disconnects after the tests pass. I _believe_ this is on our end but any feedback / guidance you have there is obviously beneficial.

Screenshot from 2019-09-27 14-13-38

test/utils/test-apis.js
daviddias: I believe this to be a misunderstanding. See https://github.com/ipfs/js-ipfs/issues/2446#issuecomment-536202616 and https://github.com/ipfs/js-ipfsd-ctl/pull/384

daviddias: The only way it would make sense is if we added support to spawn the node inside a Service/Web Worker -- https://github.com/ipfs/js-ipfsd-ctl/issues/385 -- and that would have to use https://github.com/ipfs-shipyard/ipfs-postmsg-proxy

aphelionz: Thanks David. Am I right in thinking that removing that line would make no difference?

daviddias: You are right indeed :)

achingbrain: Quite right, this line is not necessary.

It's a good idea to specify it to the `go` daemon further down though, if you are expecting the value returned from the daemon factory to have the same API as the module passed as the `exec` arg in the `proc` version, because otherwise you will get the `ipfs-http-client` bundled with the version of `ipfsd-ctl` you are using, which may not have exactly the same API as the js `ipfs` implementation.

For example `ifpsd-ctl@0.42.3` depends on `ipfs-http-client@32.0.1` but `ipfs@0.38.0-rc.6` depends on (and is only tested with) `ipfs-http-client@37.0.2` so you should keep these in sync if you want to seamlessly switch between js/go implementations as you do here.

achingbrain:
@aphelionz my experience of those sorts of socket resets is that they are requests to the daemon that are in-flight while it's being shut down, usually in an `after/afterEach` function at the end of the test suite.

aphelionz:
Merged to develop! Thanks @achingbrain. We'll be able to look at those socket issues in `develop`

achingbrain:
Great stuff. `v0.38.0` has shipped so I opened #696 to not depend on the release candidate any more.

PR #687, opened 2019-09-18
move handler to orbitdb.caches

aphelionz:


src/OrbitDB.js
```diff
@@ -48,8 +47,8 @@ class OrbitDB {
     this.storage = options.storage
     this._directConnections = {}
 
-    this.caches = { }
-    this.caches[this.directory] = options.cache
+    this.caches = {}
+    this.caches[this.directory] = { cache: options.cache, handlers: new Set() }
```
shamb0t: This poses a semantic issue no? Does it make sense that all empty-handler caches should be closed if this one is by default empty?

src/OrbitDB.js
```diff
@@ -266,34 +267,43 @@ class OrbitDB {
...
+        this.caches[this.directory].handlers.delete(address.toString())
+        if (!this.caches[this.directory].handlers.size) await this.cache.close()
       }
     }
+
     delete this.stores[address]
   }
 
+  async _onDrop (db) {
+    await db._cache.open()
```
shamb0t: Why open on drop again?

aphelionz: Because there are cache.del calls inside of Store.drop that need the cache to be open. If Store.close is called before Store.drop like in the persistency tests, this will happen.

PR #689, opened 2019-09-19
Add doc for `peer` event

phillmac:
Couldn't find any docs for https://github.com/orbitdb/orbit-db/blob/790297227706d219fe4ca92075afb20740c9b088/src/OrbitDB.js#L251
I wasn't aware such an event existed, however I think it could be an incredibly useful event in some cases, so here's some docs to match

### DIFF
--- a/API.md
+++ b/API.md
@@ -62,6 +62,7 @@ Read the **[GETTING STARTED](https://github.com/orbitdb/orbit-db/blob/master/GUI
     + [`load.progress`](#loadprogress)
   * [`ready`](#ready)
   * [`write`](#write)
+  * [`peer`](#peer)
   * [`closed`](#closed)
 
 <!-- tocstop -->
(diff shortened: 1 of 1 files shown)


API.md
```diff
@@ -603,6 +604,13 @@ db.events.on('write', (dbname, hash, entry) => ... )
 
 Emitted after an entry was added locally to the database. *hash* is the IPFS hash of the latest state of the database. *entry* is the added database op.
 
+### `peer`
+```javascript
+db.events.on('peer', (peer) => ... )
+```
+
+Emitted when a new peer connects via ipfs pubsub. *peer* is the new peer,  an instance of [`peerInfo`](https://github.com/libp2p/js-peer-info)
```
shamb0t: Is `peer` an instance of `peerInfo` or is it just the peer-id string?

shamb0t:
Good catch @phillmac and thanks for the PR! Just left one question re: the type of `peer`, is it an instance of `peerInfo` or a string?

API.md
phillmac: I didn't check back far enough into the js-ipfs implementation. Seems like its just the peer-id string.

shamb0t:
Thanks @phillmac

PR #692, opened 2019-09-20
Use store.options.directory

shamb0t:
Depends on https://github.com/orbitdb/orbit-db-store/pull/75

Simplifies `onClose` and `requestCache`


aphelionz:
:+1: LGTM. I don't think tests will pass until store is pushed

Issue #686, opened 2019-09-18
Custom Access Controller is not working


spartanz51:
Regarding the documentation here: https://github.com/orbitdb/orbit-db/blob/master/GUIDE.md#custom-access-controller

We have to implement our custom AccessController.

`class OtherAccessController extends AccessController {`

But, AccessController is not defined

So, by trying to import it:

`const AccessController = require('orbit-db-access-controllers/src/access-controller-interface')`

There is an error: 

> (node:673296) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'save' of undefined
    at Function.create (/home/anth0/Downloads/IPFSDB/node_modules/orbit-db-access-controllers/src/access-controllers.js:65:29)
    at async OrbitDB._determineAddress (/home/anth0/Downloads/IPFSDB/node_modules/orbit-db/src/OrbitDB.js:273:37)
    at async OrbitDB.create (/home/anth0/Downloads/IPFSDB/node_modules/orbit-db/src/OrbitDB.js:296:23)
(node:673296) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:673296) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.

Full code here: [https://pastebin.com/efp8xkTZ](https://pastebin.com/efp8xkTZ)

shamb0t:
Thank you @spartanz51! Apologies, the documentation is indeed out of date. A custom access controller currently needs a static `create` function which returns an instance of the access controller. Additionally a `save` method is expected which returns any parameters needed when loading an access-controller (eg saving some object to ipfs and returning the hash so that it can be retrieved on load)

The code could look sth like this https://pastebin.com/NchrdQv5 

You can see a few more examples [here](https://github.com/orbitdb/orbit-db-access-controllers/blob/master/src/contract-access-controller.js) and [here](https://github.com/orbitdb/orbit-db-access-controllers/blob/master/src/ipfs-access-controller.js). Is that helpful? 

Would you like to create a PR to fix the example in the guide to include the missing functions?

spartanz51:
Thanks for reply.
[Here](https://github.com/orbitdb/orbit-db/pull/693) is the PR

shamb0t:
[...]

PR #693, opened 2019-09-23
fix custom access controller example

spartanz51:
Fix for [#686](https://github.com/orbitdb/orbit-db/issues/686)


shamb0t:
Thank you @spartanz51!

PR #695, opened 2019-09-26
ci: Add validate-maintainers step

RichardLitt:
Testing validate-maintainers on this package. I think I've added it to the right makefile command, and this should pass on Travis if localMaintainers looks good.


RichardLitt:
\o/ It works.

RichardLitt:
Tested it with a failing test. The tests broke. That means that this should work for all of these modules. \o/

Result:

```sh
❌ There are no manually-specified npm maintainers for the `package.json` on npm.
```

One thing that may happen is that the npm package needs to have the maintainers added to it, if there are new ones specified in the PR. That is how it should be - it'll break until you remove them. So, for now, this should work for CI purposes.

RichardLitt:
> It should have logged

Scratch that. This works exactly as planned. Good.

Makefile
```diff
@@ -5,7 +5,9 @@ deps:
 
 test: deps
 	npm run test:all
-	
+	sudo npm install -g validate-maintainers
+	validate-maintainers orbit-db --match --ci
```
aphelionz: Since `test` requires `deps`, and `validate-maintainers` is in package.json's `dev-dependencies`, these two lines should be able to be compressed down to simply

```bash
npx validate-maintainers
```

RichardLitt: Thanks! You're right, this seems to work.

aphelionz:
:+1: LGTM!

PR #696, opened 2019-10-01
Upgrade to non-rc IPFS

achingbrain:
`ipfs@0.38.0` has shipped so no need to depend on the release candidate any more.


shamb0t:
Thanks @achingbrain!

PR #697, opened 2019-10-01
Feat/identities

shamb0t:
Adds an `_identities` property to orbitdb and remove it from the identity instances


test/utils/config.js
```diff
@@ -3,6 +3,9 @@ module.exports = {
   dbname: 'orbit-db-tests',
   defaultIpfsConfig: {
     start: true,
+    EXPERIMENTAL: {
```
aphelionz: Do we still need this, post js-ipfs 0.38?

shamb0t:
Closing in favor of keeping the `provider` reference in `identity` so that they are not managed separately

PR #698, opened 2019-10-02
Unit tests that verify calling load() on a feed with an amount greater than the number of items loads all the items

ptoner:
These tests create a feed and add 10 records to it. Then it tries to call load() with various amounts that are greater than 10.

### DIFF
--- a/test/feed-load.test.js
+++ b/test/feed-load.test.js
@@ -1,0 +1,108 @@
+'use strict'
+
+const assert = require('assert')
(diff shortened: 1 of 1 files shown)


shamb0t:
Thanks @ptoner ! Indeed looks like the amount needs to be clamped

shamb0t:
Tests now passing with fix in https://github.com/orbitdb/ipfs-log/pull/270, thank you @ptoner!

PR #703, opened 2019-10-08
add field-manual to readme

shamb0t:
Adds link to field-manual in usage section of readme


sirfumblestone:
LGTM 👍 

Should we update the license from MIT © 2015-2018 to 2015-2019 on the same go?

PR #712, opened 2019-11-03
Add Windows support.

kolessios:
This PR adds Windows support with the least possible number of changes, unlike other PRs with the same objective, this changes try to be friendly with the current code. 

This is a summary of the most important changes:

- The libraries `cpy-cli` and `mkdirp` were added to be alternatives to the `cp` and `mkdir` commands found in the `package.json` scripts. [View](https://github.com/orbitdb/orbit-db/compare/master...dreamnettech:feature/windows-support?expand=1#diff-b9cfc7f2cdf78a7f4b91a753d10865a2)
- The `cross-env` library was added to execute the `test` command on all platforms.
- The areas affected by `path.join` (orbit addresses) were changed to `(path.posix || path)`. posix provides the correct way to join the strings and if it does not exist (in case of use on website) the normal `path` is used.
- A static method called `join` has been created in the `OrbitDBAddress` class that allows to replace `path.join` for the purpose of the previous point. [View](https://github.com/orbitdb/orbit-db/compare/master...dreamnettech:feature/windows-support?expand=1#diff-20abc58686fca870e14cd87332aac208R55)
- Windows has problems copying the `LOG` and `LOCK` files from `tests/fixtures`, a small filter has been created to avoid copying these files. [View](https://github.com/orbitdb/orbit-db/compare/master...dreamnettech:feature/windows-support?expand=1#diff-d526dfb96df62ac12de4880998ae62deR43)
- Windows has trouble reading the files/binaries from `tests/fixtures` when they have been saved with **EOL: CRLF**, this causes 4 tests to fail, the solution is to prevent git from changing the EOL of the files. [Source](https://github.com/ipfs/js-ipfs-repo/pull/153#issuecomment-346320309)
- 2 tests have been added to validate that an address with backslashes is considered valid.  [Source](https://github.com/orbitdb/orbit-db/pull/453/files#diff-6459cdfcb4da790fc82c55906ebce4c5)
- ⚠ For some unknown reason the `load v0 orbitdb address` test fails most of the time (but not always) when it is first executed with the `test:all` command, if the test is run individually it passes without problems. I have called `db.load()` before each test to correct this problem, however it is not a real solution.
- Other changes in the code are the result of running [standard](https://standardjs.com/)

---

- All tests run smoothly in:
  - Windows 10 1903 (Build 18362)
  - Node v10.16.2


dacom-dark-sun:
Excellent! Hope, it will be accepted by a core team.

aphelionz:
Is there anything standing between us and merging this? @haadcode @shamb0t ?

shamb0t:
Thanks a lot for the summary, @kolessios! Trying on a windows machine and the only failing test is the v0-backwards compatibility due to identity-migration failing, which is expected, so this looks good to merge :+1:

PR #713, opened 2019-11-05
Add db events to orbitdb instance

phillmac:
Add an 'open' event, fired when a store has completed opening.
Useful in situations where it could take an extended period of time for a store to open and instead of awaiting `OrbitDB.open()` an event listener can be used.
Likewise it would be useful to be notified of 'ready' events without having first subscribed to a particular db intance. May as well go the whole hog and implement all the events emitted by dbs.


phillmac:
Replaced by #793

Issue #715, opened 2019-11-06
Expose modules in OrbitDB class for the browser


hazae41:
It's ridiculous that we need to reimport these modules while OrbitDB actually uses them.

How about static getters? so we can use

    const { OrbitDB } = window;
    const { Store } = OrbitDB;
    const { AccessControllers, AccessController } = OrbitDB;
    const { Identities, IdentityProvider } = OrbitDB;
    ...

Also, we should be able to add a custom access controller without needing the AccessControllers object

    OrbitDB.addAccessController(MyAccessController)

Same with identity providers

    OrbitDB.addIdentityProvider(MyIdentityProvider)
    
While we're at it, `OrbitDB.addDatabaseType(CustomStore.type, CustomStore)` should be simplified to  `OrbitDB.addDatabaseType(CustomStore)` and get the type directly from `CustomType.type`

aphelionz:
While "ridiculous" might be subjective and debatable, I do agree that it would be convenient to have these modules available for custom work without having to explicitly import, for example, `orbit-db-access-controllers`

An easy way to achieve this might be to simply add them to the prototype of the already-exported OrbitDB class.

This would maintain backwards compatibility with existing codebases that are still using `const OrbitDB = require('orbit-db')` and still allow you the ease of customization via accessors like `OrbitDB.AccessControllers`

shamb0t:
This was resolved by #717

PR #717, opened 2019-11-06
Expose modules in OrbitDB object

hazae41:
so we can grab them from OrbitDB without reimporting them.

Related to https://github.com/orbitdb/orbit-db/issues/715


phillmac:
IMHO it'd be really useful if we could add the OrbitDBAddress class to this. Many a time I've wanted to check an address is valid but haven't wanted to import that  class

PR #720, opened 2019-11-08
Error if address path mismatches manifest name

phillmac:
Currently it's possible to open a db via an address with a different path component then what the original manifest name was set to. This Pr prevents the issue by throwing an error is such a mismatch occurs.


shamb0t:
Thanks @phillmac !

PR #722, opened 2019-11-11
Update README.md

hazae41:
pubsub enabled by default

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,18 +90,9 @@ npm install orbit-db ipfs
 ```javascript
 const IPFS = require('ipfs')
 const OrbitDB = require('orbit-db')
-// OrbitDB uses Pubsub which is an experimental feature
-// and need to be turned on manually.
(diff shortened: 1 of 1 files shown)


aphelionz:
Looks good but we probably want to leave a note that says if people are using pre 0.39 then they will still need to enable pubsub.

README.md
```diff
@@ -90,17 +90,18 @@ npm install orbit-db ipfs
...
-// Note that these options need to be passed to IPFS in
-// all examples even if not specified so.
+
+// Create IPFS instance
+
+// For js-ipfs >= 0.38
+const ipfs = new IPFS()
+
+// For js-ipfs < 0.38
 const ipfsOptions = {
```
RichardLitt: I don't think these four lines are needed anymore.

aphelionz: I specifically requested them :D

RichardLitt: Ah. I see the changes. Nevermind! :D

PR #725, opened 2019-11-14
Catch unhandled error in test

haadcode:
This PR will catch an unhandled (but expected) error in one of the tests. 

Before, the error was seen in the test output. After this PR, the error is not shown in the output (expected).


shamb0t:
Thanks!

Issue #752, opened 2020-01-07
Share heads from cache on peer connected


zachferland:
Possibly to speed up replication, share heads on peer join from cache when available vs waiting for the log to be loaded. Peer only needs heads to start syncing instead of waiting for other peer to finish loading db. 

More relevant probably to pinning services (3box for example), where when a db wants to pin the log or sync it tells the pinning service to open that db. Then peer has to wait for that service to load and open log. 

Also related to https://github.com/orbitdb/orbit-db/pull/688 where the heads are shared on connecting, but they are sometimes null (and not loaded yet). So we want to share heads on connect still, but make sure available at that time (from cache instead of log) (pr also good reference for which lines of code this references)

implementing in 3box first, and will likely share pr here soon as well

zachferland:
Thanks for quickly getting it into a release! closing issue now

PR #754, opened 2020-01-09
feat: share heads from cache on peer connect, available before log ready

zachferland:
Implements https://github.com/orbitdb/orbit-db/issues/752

Also partly resolves this fix pr i wrote earlier, with this implementation heads will not longer be [ ] if log is not ready at time of onpeer, and they can still be shared on connecting (instead of waiting)
https://github.com/orbitdb/orbit-db/pull/688


shamb0t:
Thanks @zachferland LGTM :+1:

PR #756, opened 2020-01-14
adds toJSON method to orbitdb address class

tabcat:
Currently json serializing an orbitdb address results in the json value `{ root: "<root>", path: "<path>" }` which throws an error when fed back into address.parse.

Adding a toJSON method that returns this.toString() would make the json serialized address again parseable by the address class.

This is small change but is breaking the behavior of how json.stringify serializes the address currently.

### DIFF
--- a/src/orbit-db-address.js
+++ b/src/orbit-db-address.js
@@ -14,6 +14,10 @@ class OrbitDBAddress {
     return OrbitDBAddress.join(this.root, this.path)
   }
 
+  toJSON () {
+    return this.toString()
+  }
+
   static isValid (address) {
     address = address.toString().replace(/\\/g, '/')
 
--- a/test/orbit-db-address.test.js
+++ b/test/orbit-db-address.test.js
@@ -63,6 +63,24 @@ Object.keys(testAPIs).forEach(API => {
         assert.equal(result.toString().indexOf('zd'), 9)
       })
 
+      it('parse address after json serialization', () => {
+        const address = JSON.parse(JSON.stringify(
+          OrbitDB.parseAddress(
+            '/orbitdb/zdpuAuK3BHpS7NvMBivynypqciYCuy2UW77XYBPUYRnLjnw13/first-database'
+          )
+        ))
+        const result = OrbitDB.parseAddress(address)
+
+        const isInstanceOf = result instanceof OrbitDBAddress
+        assert.equal(isInstanceOf, true)
+
+        assert.equal(result.root, 'zdpuAuK3BHpS7NvMBivynypqciYCuy2UW77XYBPUYRnLjnw13')
+        assert.equal(result.path, 'first-database')
+
+        assert.equal(result.toString().indexOf('/orbitdb'), 0)
+        assert.equal(result.toString().indexOf('zd'), 9)
+      })
+
       it('parse address with backslashes (win32) successfully', () => {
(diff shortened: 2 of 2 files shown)


tabcat:
any reviewers?

src/orbit-db-address.js
```diff
@@ -14,6 +14,10 @@ class OrbitDBAddress {
     return OrbitDBAddress.join(this.root, this.path)
   }
 
+  toJSON () {
```
aphelionz: I'm afraid having a toJSON method that returns a string makes me a little dizzy. Can we check the type in `address.parse` and handle it accordingly there?

tabcat: Thank you for taking time to look at this.
I thought about changing the parser initially when wanting to make this pr but decided not to.
I am opposed to only changing the parser and not having toJSON return a string.
One reason for this is I think orbitdb addresses are better described/more readable as strings. They also resemble multiaddrs a bit which have a toJSON method that return a string here: https://github.com/multiformats/js-multiaddr/blob/072326e4bc9a0fa5fab0316d8160d2b493534e8f/src/index.js#L69
It seems like most things in javascript that have a toString method would also be a string in json.
I'm not opposed to changing the parser to handle the current json form of the address but I think its unnecessary.

aphelionz: Thanks for making the PR!

I suppose one could make the the argument that a string in JS could feasibly be viewed as "valid JSON", I still don't like the idea of messing with expectations in this way. JS types are more are less free-for all, but I'd like to honor them as much as possible... Also, if `JSON.parse("xxxx")` didn't throw an error and returned a string I might be more persuaded.

Other people are free to weigh in (maybe toss it out to Gitter?) but for better or worse I'm still on the side of moving this handler to the parse function. I'm sorry if this is frustrating.

tabcat: Its not frustrating, I appreciate the honest feedback. <3

tabcat: it is possible to parse json strings like so `JSON.parse('"asdf"')`

aphelionz: I'm sorry I didn't ask this before - what exactly is the error you're getting and on which line?

tabcat:
```
address.constructor.parse(
  JSON.parse(JSON.stringify(address))
)

resulting error:
orbit-db-address.js:42 Uncaught Error: Not a valid OrbitDB address: [object Object]
    at Function.parse (orbit-db-address.js:42)
    at <anonymous>:1:18
```
https://github.com/orbitdb/orbit-db/blob/ed697fdfd6bead8752d6d0a4760270653d3215a8/src/orbit-db-address.js#L30
here `parts` is equal to `["[object Object]"]` which makes `accessControllerHash` conditional false.

aphelionz: Can we start by writing a failing test to catch this case? And then I can play around with it a little bit on my own and provide more cohesive thoughts.

tabcat: I did include a test that sounds like what you want, but if its not lmk.

aphelionz: Ah, so you did! I'll take a look at this after I've had some more sleep :)

aphelionz:
Correct me if I'm wrong. I think this can be closed.

PR #763, opened 2020-02-19
Update peer exchanged

msterle:
Bring https://github.com/orbitdb/orbit-db/pull/606 up to date


msterle:
Interestingly, these tests pass when running locally, even without the updated `orbit-db-store` package, but fail here on the orbit-db CI. It's possible that this is the sort of timing condition that the `peer.exchanged` test was trying to capture, but I'll have to look into it more - ideally, the tests should pass or fail consistently, independent of environment.

aphelionz:
@msterle Looks like we may have IPFS 0.41 support soon... that could will help.

msterle:
Test fixed by increasing the time between the final `replicate` event received and the test for number of `peer.exchanged` events.

aphelionz:
I'm curious now if the update from #782 will now fix this! Let's see if we can get it into 0.24.0.

aphelionz:
@msterle would you mind rebasing this to the latest master? Super curious to see how the tests do

aphelionz:
@oed I don't think so!

oed:
Updated @aphelionz, looks like tests are failing on master as well.

aphelionz:
@oed @msterle Any chance of getting this rebased one more time now that we have passing tests on master?

oed:
Done @aphelionz

PR #780, opened 2020-04-17
add API to GUIDE.md

tabcat:
creates a field in the guide doc that is very similar to what is on the main readme of the repo
test it out here: https://github.com/tabcat/orbit-db/blob/patch-1/GUIDE.md#api


aphelionz:
LGTM, will merge when green

PR #767, opened 2020-03-06
ready for js-ipfs >= 0.41

Xmader:
close #762

PR #782, opened 2020-04-21
fix: support POST-only API of go-ipfs 0.5

lidel:
## TL;DR

This PR ensures orbit-db remains compatible with HTTP API exposed by go-ipfs 0.5 (https://github.com/ipfs/go-ipfs/issues/7109) by switching to `ipfs-http-client@41.0.1` in tests and adding interop note in README.  

## Changes

- **added interop note to README about compatible http client version**
  - note about version limitation can be removed when #767 lands
- bumped versions of http-client and go-ipfs used in tests to ensure it works with latest go-ipfs 0.5.0-rc2

## Why?

go-ipfs 0.5  will block `GET` commands on the API port (https://github.com/ipfs/go-ipfs/pull/7097), requiring every  command (RPC) to be sent as HTTP `POST` request.

Why v41.0.1? It sends only POST requests while maintaining the old JS API 
(breaking changes that introduced support for async iterators landed in 42.x)

cc @aphelionz @stebalien @achingbrain

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,13 +160,16 @@ initIPFSInstance().then(ipfs => {
 ```
 
 ### Module with IPFS Daemon
-Alternatively, you can use [ipfs-api](https://npmjs.org/package/ipfs-api) to use `orbit-db` with a locally running IPFS daemon. Use this method if you're using `orbitd-db` to develop **backend** or **desktop** applications, eg. with [Electron](https://electron.atom.io).
+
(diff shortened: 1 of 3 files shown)


aphelionz:
Thank you @lidel!!

Stebalien:
The go-ipfs 0.5.0 release will be cut later today. Any chance this can be merged soon?

aphelionz:
@Stebalien Merged to master for now, before our own release. Hope that's helpful!

Stebalien:
It does. Thanks!

aphelionz:
@lidel @Stebalien should the practice org-wide in `orbitdb` be for us to utilize `ipfs-http-client@~41.0.1` and `go-ipfs@0.5.0-rc2`?

lidel:
Yes (0.5.0 just shipped) :)

Stebalien:
Is there any chance we could cut a patch release with this change so users can upgrade go-ipfs without migrating to async/await?

PR #783, opened 2020-04-22
fix: fs shim when running inside electron renderer process

mistakia:
The fs shim does not work properly when used inside an [electron renderer process](https://www.electronjs.org/docs/tutorial/application-architecture#main-and-renderer-processes), where both native node modules and chromium (i.e. global `window`) are available. This is one simple way to address the issue.


aphelionz:
Thanks! We can get this into 0.24.0

PR #786, opened 2020-05-04
add meta option to doc

tabcat:
This documents the meta option available at db creation. Includes input type expected, the default for the option, and a description of what it does/affects.

Looking for feedback on wording especially.

### DIFF
--- a/API.md
+++ b/API.md
@@ -115,6 +115,8 @@ Returns a `Promise` that resolves to [a database instance](#store-api). `name` (
 
 - `overwrite` (boolean): Overwrite an existing database (Default: `false`)
 - `replicate` (boolean): Replicate the database with peers, requires IPFS PubSub. (Default: `true`)
+- `meta` (object): An optional object in [database manifest](https://github.com/orbitdb/orbit-db/blob/master/GUIDE.md#address). Immutably stores any JSON-serializable value. Readable via `db.options.meta`. Default: `undefined`.
+
 ```javascript
 const db = await orbitdb.create('user.posts', 'eventlog', {
     accessController: {
@@ -124,7 +126,10 @@ const db = await orbitdb.create('user.posts', 'eventlog', {
         // Give access to the second peer
         '042c07044e7ea51a489c02854db5e09f0191690dc59db0afd95328c9db614a2976e088cab7c86d7e48183191258fc59dc699653508ce25bf0369d67f33d5d77839'
       ]
-    }
+    },
+    overwrite: true,
+    replicate: false,
+    meta: { hello: 'meta hello' }
 })
 // db created & opened
 ```


API.md
```diff
@@ -115,6 +115,8 @@ Returns a `Promise` that resolves to [a database instance](#store-api). `name` (
 
 - `overwrite` (boolean): Overwrite an existing database (Default: `false`)
 - `replicate` (boolean): Replicate the database with peers, requires IPFS PubSub. (Default: `true`)
+- `meta` (object): An object to be added to the database manifest (learn more about the [db manifest](https://github.com/orbitdb/orbit-db/blob/master/GUIDE.md#address)). Data in the manifest affects the address of the db and cannot be changed after creation. It's recommended to store smaller objects and must be serializable to json. Read the object at instance property `db.options.meta`. Default: `undefined`.
+
 ```javascript
```
aphelionz, line 120: I know this isn't your job... In this PR but would you mind adding all of the above options to the JS example for posterity?

API.md
```diff
@@ -115,6 +115,8 @@ Returns a `Promise` that resolves to [a database instance](#store-api). `name` (
 
 - `overwrite` (boolean): Overwrite an existing database (Default: `false`)
 - `replicate` (boolean): Replicate the database with peers, requires IPFS PubSub. (Default: `true`)
+- `meta` (object): An object to be added to the database manifest (learn more about the [db manifest](https://github.com/orbitdb/orbit-db/blob/master/GUIDE.md#address)). Data in the manifest affects the address of the db and cannot be changed after creation. It's recommended to store smaller objects and must be serializable to json. Read the object at instance property `db.options.meta`. Default: `undefined`.
```
aphelionz, line 118: How about:

- `meta` (object): An optional object [database manifest](https://github.com/orbitdb/orbit-db/blob/master/GUIDE.md#address)). Immutably stores any JSON-serializable key/value pair and outputs via `db.options.meta`. Default: `undefined`.

aphelionz: There's also the idea that these things will affect the DB address but I'm not sure if here specifically is the place to put that info

aphelionz:
ha, whoops. Why close?

tabcat:
I was worried that something else was included with this but it wasnt.

API.md
tabcat: sure, should this be in a separate pr?

aphelionz: I wouldn't mind if it was in the same PR since it's doc-only

API.md
tabcat: much more concise which is nice, i would slightly rather `key/value pair` -> `value` and maybe `outputs` -> `readable`?

aphelionz: yeah that's fine

tabcat:
any last minute changes i should make?

aphelionz:
I'm gonna merge this despite the CI errors since you clearly didn't introduce them here

PR #793, opened 2020-05-12
Relay events from db scope to orbitdb scope

phillmac:
Adds a `realyEvents:[]` param to :
 - ` OrbitDB.createInstance()`
 - `orbit.db.create()`
 - `orbitdb.open()`

This enables listeners to be created at the parent orbitdb instance level, allowing events to be trapped for db creation before the creation process even begins.
This is especially needed for the http api where opening a remote db can take extended periods of time, and http clients will enforce a relatively shorter time-out period when there is no activity.
Some basic tests are included for the `open`, `load` and `ready` events
More tests can be added later if needed.
 
Example:
```
orbitdb.events.on('ready', ((addr) => console.log(`DB ${addr} is ready`))
db = await orbitdb.open('ready_event_test', { create: true, type: 'feed', relayEvents:['ready'] })
```

### DIFF
--- a/src/OrbitDB.js
+++ b/src/OrbitDB.js
@@ -238,28 +238,25 @@ class OrbitDB {
       onDrop: this._onDrop.bind(this),
       onLoad: this._onLoad.bind(this)
     })
-
-    const relayEvents = options.relayEvents ? options.relayEvents : this._relayEvents
-
     const identity = options.identity || this.identity
 
     const store = new Store(this._ipfs, identity, address, opts)
     store.events.on('write', this._onWrite.bind(this))
 
-    const eventRelay = this.events
-    for (const eventType of relayEvents) {
-      store.events.on(eventType, (...args) => eventRelay.emit(eventType, address, ...args))
-    }
-
     // ID of the store is the address as a string
     const addr = address.toString()
     this.stores[addr] = store
+
+    const relayEvents = options.relayEvents || this._relayEvents
+    for (const eventType of relayEvents) {
+      store.events.on(eventType, (...args) => this.events.emit(eventType, addr, ...args))
+    }
 
     // Subscribe to pubsub to get updates from peers,
     // this is what hooks us into the message propagation layer
     // and the p2p network
     if (opts.replicate && this._pubsub) { await this._pubsub.subscribe(addr, this._onMessage.bind(this), this._onPeerConnected.bind(this)) }
-    this.events.emit('open', store.address.toString())
+    this.events.emit('open', addr)
     return store
   }
 
--- a/src/OrbitDB.js
+++ b/src/OrbitDB.js
@@ -339,7 +339,6 @@ class OrbitDB {
     const dir = db && db.options.directory ? db.options.directory : this.directory
     await this._requestCache(address, dir, db._cache)
     this.stores[address] = db
-    this.events.emit('load', address)
   }
 
   async _determineAddress (name, type, options = {}) {
--- a/src/OrbitDB.js
+++ b/src/OrbitDB.js
@@ -17,6 +17,7 @@ const exchangeHeads = require('./exchange-heads')
 const { isDefined, io } = require('./utils')
 const Storage = require('orbit-db-storage-adapter')
 const migrations = require('./migrations')
+const { EventEmitter } = require('events')
 
 const Logger = require('logplease')
 const logger = Logger.create('orbit-db')
@@ -53,6 +54,8 @@ class OrbitDB {
     this.caches[this.directory] = { cache: options.cache, handlers: new Set() }
     this.keystore = options.keystore
     this.stores = {}
+    this._relayEvents = options.relayEvents
+    this.events = new EventEmitter()
 
     // AccessControllers module can be passed in to enable
     // testing with orbit-db-access-controller
@@ -118,6 +121,10 @@ class OrbitDB {
       const cachePath = path.join(options.directory, id, '/cache')
       const cacheStorage = await options.storage.createStore(cachePath)
       options.cache = new Cache(cacheStorage)
+    }
+
+    if (!options.relayEvents) {
+      options.relayEvents = []
     }
 
     const finalOptions = Object.assign({}, options, { peerId: id })
@@ -195,6 +202,11 @@ class OrbitDB {
 
     // Remove all databases from the state
     this.stores = {}
+
+    // Remove all event listeners
+    for (var event in this.events._events) {
+      this.events.removeAllListeners(event)
+    }
   }
 
   // Alias for disconnect()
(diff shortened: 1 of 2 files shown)


src/OrbitDB.js
```diff
@@ -226,11 +238,19 @@ class OrbitDB {
       onDrop: this._onDrop.bind(this),
       onLoad: this._onLoad.bind(this)
     })
+
+    const relayEvents = options.relayEvents ? options.relayEvents : this._relayEvents
```
tabcat, line 242: can be reduced to `const relayEvents = options.relayEvents || this._relayEvents` .
personally i would also move this closer to where it's used. i see you are trying to keep it by all the other option related stuffs.

src/OrbitDB.js
```diff
@@ -226,11 +238,19 @@ class OrbitDB {
...
+
     const identity = options.identity || this.identity
 
     const store = new Store(this._ipfs, identity, address, opts)
     store.events.on('write', this._onWrite.bind(this))
 
+    const eventRelay = this.events
```
tabcat, line 249: what is the purpose of declaring `eventRelay`

src/OrbitDB.js
```diff
@@ -239,7 +259,7 @@ class OrbitDB {
     // this is what hooks us into the message propagation layer
     // and the p2p network
     if (opts.replicate && this._pubsub) { await this._pubsub.subscribe(addr, this._onMessage.bind(this), this._onPeerConnected.bind(this)) }
-
+    this.events.emit('open', store.address.toString())
```
tabcat, line 262: I feel like this is the meat of the pr and allows for users to do everything else with minimal logic.

src/OrbitDB.js
phillmac: so I don't have to use `this` in the arrow function below it, `this` goes out of in the closure, so a separate reference to `this.events` is required.

src/OrbitDB.js
phillmac: Yeah, true. Mostly I want just these three events, `open`, `load`, and `ready`. `ready` comes from the db scope so it's better imo to just add the facility to hoist any and all db scope events to the parent instance scope

src/OrbitDB.js
tabcat: what would you say `this` is referring to when inside of the for loop statements brackets?

src/OrbitDB.js
phillmac: Done!

src/OrbitDB.js
phillmac: Inside the for-loop it refers to the orbitdb instance.
However, inside the closure inside the for loop, the arrow function, it refers to the context of the arrow function

tabcat: arrow functions have lexical scope so you should be good here

phillmac: My bad, arrow functions don't change scope like ordinary functions do.
example: 
```
const { EventEmitter } = require('events')
(function () {
  const address = 'zdpu.....ect'
  this.events = new EventEmitter()
  const store = { events: new EventEmitter() }
  const relayEvents = ['ready']

  for (const eventType of relayEvents) {
    store.events.on(eventType, function (...args) {
      console.log(typeof this.events) //undefined
      this.events.emit(eventType, address, ...args) //TypeError: Cannot read property 'emit' of undefined
    })
  }

  this.events.on('ready', (...args) => console.log(`Got ready event with args ${args}`))

  store.events.emit('ready', 'this is a test')
}) ()
```

phillmac: So it turns out that arrow functions don't change the scope of this like standard functions do.
e.g.
```
const { EventEmitter } = require('events')

function abc () {
  const address = 'zdpu.....ect'
  this.events = new EventEmitter()
  const store = { events: new EventEmitter() }
  const relayEvents = ['ready']

  for (const eventType of relayEvents) {
    store.events.on(eventType, function (...args) {
      console.log(typeof this.events)                       //undefined
      this.events.emit(eventType, address, ...args)} //TypeError: Cannot read property 'emit' of undefined
    )
  }

  this.events.on('ready',(...args) => console.log(`Got ready event with args ${args}`))

  store.events.emit('ready', 'this is a test')
}
abc()
``` 

e.g. 2 works fine
```
const { EventEmitter } = require('events')

function abc () {
  const address = 'zdpu.....ect'
  this.events = new EventEmitter()
  const store = { events: new EventEmitter() }
  const relayEvents = ['ready']

  for (const eventType of relayEvents) {
    store.events.on(eventType, (...args) => {
      console.log(typeof this.events)                          //object
      this.events.emit(eventType, address, ...args)}
    )
  }

  this.events.on('ready',(...args) => console.log(`Got ready event with args ${args}`))

  store.events.emit('ready', 'this is a test')
}
abc()
```

phillmac: Fixed!

tabcat:
I'm mainly interested in adding the 'open' event. It it enables users to attach their own listeners directly when a new store is opened. Although I'm not convinced even that would be very beneficial.

src/OrbitDB.js
```diff
@@ -319,6 +336,7 @@ class OrbitDB {
     const dir = db && db.options.directory ? db.options.directory : this.directory
     await this._requestCache(address, dir, db._cache)
     this.stores[address] = db
+    this.events.emit('load', address)
```
tabcat, line 339: i believe if a relayEvent of 'load' existed then two load events would fire from this.events for the same store.

phillmac: If I remove this line and update the test like so: 
```
it('emits load event when opening database', async () => {
          let dbAddres
          const handleEvent = (addr => dbAddres = addr)
          orbitdb.events.once('load',  handleEvent)
          db = await orbitdb.open('load_event_test', { create: true, type: 'feed', relayEvents:['load'] })
          await db.load()
          assert.equal(dbAddres, db.address.toString())
      })
```

it results in a test failure:
```
1) orbit-db - Create & Open (js-ipfs)
       Open
         emits load event when opening database:
     AssertionError [ERR_ASSERTION]: undefined == '/orbitdb/zdpuAvZRphzghitg8ZbJY3hx5UeLEKPWNLam3koaMkKhKMZPZ/load_event_test'
      at Context.<anonymous> (test/create-open.test.js:397:18)
```
What you said should be true, yet it isn't. there's a flaw somewhere, but I'm not able to see if :C

phillmac:
```
 if (heads.length > 0) {
      this.events.emit('load', this.address.toString(), heads)
    }
```
So load event only fires if heads is greater then zero..... idk what to do here

phillmac: Removed it for now, and fixed the tests, this may need to be revisited in the future.

aphelionz:
@phillmac refresh my memory here? :D

phillmac:
Basically this was someting I needed to fox http timeout errors with clients for the http api. It would take too long to open a db, so the client would spit an error and go away. Couldn't listen to events on a db that doesn't exist yet, so it seemed like a good idea to try sort it out in one go. This was pretty much ready to merge except for the "memory leak" too many listeners open that was happening elsewhere, and wanted to fix that first before adding anything new

aphelionz:
Ah yes... the event listener leak thing...

phillmac:
Incidentally, did you everf fix that? I ran into a situation where it became really bad because I had missing ennltries that were not available on any peers. It seemed like a bunch of callbacks in the replicstion progress handler were never being cmpleted

aphelionz:
Not yet, that's next on the list because we need to refactor the replicator anyway

PR #794, opened 2020-05-13
wip: updating tests

aphelionz:


aphelionz:
Thank you @shamb0t!!

aphelionz:
Still not quite sure what to do about these:

```
(node:32737) MaxListenersExceededWarning: Possible EventEmitter memory leak detected.
 11 replicate.progress listeners added to [EventEmitter]. Use emitter.setMaxListeners
() to increase limit
```

Obviously we could follow the suggestion and boost the maxListeners but I'd rather not do that without knowing exactly what's going on and if its a problem with the tests vs the library

test/counterdb.test.js
```diff
@@ -127,9 +127,12 @@ Object.keys(testAPIs).forEach(API => {
 
         return new Promise(resolve => {
           // Wait for a while to make sure db's have been synced
-          setTimeout(() => {
+          setTimeout(async () => {
```
tabcat, line 130: could awaiting replicated events be good here?

aphelionz: Those particular setTimeouts can actually go, since I moved that shim to [`orbit-db-test-utils`](https://github.com/orbitdb/orbit-db-test-utils/blob/master/stop-ipfs.js#L13)

As far as awaiting the replicated events, that could possibly help...

test/custom-cache.test.js
```diff
@@ -41,25 +44,24 @@ Object.keys(testAPIs).forEach(API => {
...
-      if(orbitdb1)
+    after(() => {
+      // FIXME: shim to wait until the end of the event loop to call this
+      setTimeout(async () => {
         await orbitdb1.stop()
-
-      if (ipfsd)
         await stopIpfs(ipfsd)
+      }, 0)
     })
```
tabcat, line 53: what is this doing exactly. the function handed to after is completing before anything is actually done.
im sure theres a reason for this and im curious.

test/counterdb.test.js
tabcat: if the replicated eventListeners arent needed they should be left out.

test/custom-cache.test.js
aphelionz: adding a `setTimeout` with a time of 0 just makes sure the inner function is executed at the end of the event loop i.e. after all eventListeners are fired

tabcat: is this good practice for mocha tests or are there specific eventListeners you want to make sure are done?

aphelionz: It's a hack, but likely a necessary one since I believe these eventListeners are at the IPFS level and not necessarily at the OrbitDB level

test/counterdb.test.js
aphelionz: Sorry, where are you seeing the replicated event listeners?

tabcat: Oh sorry. I mean if they aren't needed to make sure that both counters are utd, they don't need to be added.

test/custom-cache.test.js
aphelionz: Either way I'll remove the extraneous `setTimeout` calls at the `orbit-db` level since `orbit-db-test-utils` handles it

aphelionz:
Thanks @achingbrain ! While you're here might you have any insight into how to properly debug and handle the `EventEmitter` issues, based on your work with `js-ipfs`?

aphelionz:
Working on this. Here's what's left before I'm comfortable releasing 0.24.0:

[ ] Replication Performance
[ ] Keystore / Docstore Performance
[ ] EventEmitters going out of control and/or increasing maxEventListeners

@oed might you be able to push out a @next npm tag of 3box so the folks in #767 can move forward?

oed:
@aphelionz Are those things problems that have appeared when updating to the latest version of ipfs, or are they just longer running issues? 
If it's the second it would be nice to have a release that just featured the ipfs update, but understand if not. 

And yeah we are working on updating ipfs this and next week, so we may do a `next` release before that.

aphelionz:
@oed Alright, if that's the case then let me just get the release out as-is and then we can patch. People have been reporting that 0.24.0 has been working for them

oed:
That would be really helpful @aphelionz :pray:

PR #799, opened 2020-05-27
chore: update go-ipfs to 0.5.1

Stebalien:
It was locked at 0.5.0-rc2.


phillmac:
What's this do? Seems kinda funky to have a `go-ipfs` that's not attached to either dependencies or dev-dependencies?

aphelionz:
@phillmac It's part of `npm-go-ipfs-dep`

PR #804, opened 2020-06-24
Suport new base36 format, Improve unwieldy logic

phillmac:
The 0.6.0 version of go-ipfs introduces base36 cids, so It'd be good if we can get ahead of any issues that might occur when (if) this becomes the default and support it right out of the gate. The logic with the existing single if statement was rapidly getting out of hand and foreseeably was going to get worse over time given the amount of format changes cids are undergoing,; So it required a small refactor to make it neater & easier to add new cases.
e.g. base36 cid `k51qzi5uqu5dj16qyiq0tajolkojyl9qdkr254920wxv7ghtuwcz593tp69z9m`

### DIFF
--- a/src/orbit-db-address.js
+++ b/src/orbit-db-address.js
@@ -26,8 +26,18 @@ class OrbitDBAddress {
 
     let accessControllerHash
 
+    const validateHash = (hash) => {
+      const prefixes = ['zd', 'Qm', 'ba', 'k5']
+      for (const p of prefixes) {
+        if (hash.indexOf(p) > -1) {
+          return true
+        }
+      }
+      return false
+    }
+
     try {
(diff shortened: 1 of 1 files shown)


haadcode:
Sorry @phillmac it's taken so long to get to this. It look good to me :+1:

Thank you! 🙏

aphelionz:
Should we publish a patch release on 0.24?

phillmac:
> Sorry @phillmac it's taken so long to get to this. It look good to me
Np there was no rush :grin:

> Should we publish a patch release on 0.24?
Only if someone needs base36 cids support urgently. I've not even heard a peep outta anyone needing this yet. I just remember we got blindsided with base32 cids last time and I thought It'd be good if we could jump on this _before_ it turns into a problem.
In any case if we intend to publish 0.25 in the next few months I think that'll be fine.

PR #837, opened 2020-09-05
Updating for js-ipfs 0.50 and go-ipfs 0.6.0

aphelionz:
TODO:

- [x] ~Update the `ipfs-pubsub-*` repos one way or another~
- [x] Update the `orbit-db-****store` repos `peerDependency` field as they are in master (:tada: the first of the Roadmap items to be done!)
- [x] https://github.com/orbitdb/orbit-db-store/pull/98

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,7 @@ jobs:
     # To see the list of pre-built images that CircleCI provides for most common languages see
     # https://circleci.com/docs/2.0/circleci-images/
     docker:
-      - image: circleci/node:10-browsers
+      - image: circleci/node:12-browsers
     steps:
     # Machine Setup
     #   If you break your build into multiple jobs with workflows, you will probably want to do the parts of this that are relevant in each
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ deps:
 	npm install
 
 test: deps
-	npm run test:all
+	npm run test:all -- --exit
 	npx validate-maintainers orbit-db --match --ci
 
 build: test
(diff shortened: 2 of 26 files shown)


aphelionz:
Status: This is looking pretty alright, except for the timeouts in go-ipfs in these tests:

image

Moving some timing bits around in the tests (opening the db before starting to write entries to them) helps, but then we get lots of errors in `ipfs-pubsub-1on1` and `ipfs-pubsub-peer-monitor` that break the tests and prevent cleanup from happening.

I opened issues in those repos asking for access or to just move them here. :crossed_fingers: that would make working on those and cleaning up those timers a lot easier.

- https://github.com/ipfs-shipyard/ipfs-pubsub-1on1/issues/14
- https://github.com/ipfs-shipyard/ipfs-pubsub-peer-monitor/issues/2

aphelionz:
So this is working - I need to move on to other stuff besides debugging the `ipfs-pubsub-*` repos. So that leaves us with a choice - should we publish 0.26.0 and deal with the tests hanging periodically or wait and deal with it sometime next week (or after?)

aphelionz:
cc @oed @tabcat @vaultec81 @vasa-develop @haadcode

tabcat:
If I understand correctly either specific tests to do with replication are timing out or tests arent exiting, and either behavior is inconsistent. Leaving it published as a release candidate cant hurt. If specific replication tests are timing out often its not out of the ordinary in my experience, but this sounds more like the other behavior (tests not exiting).

aphelionz:
Yes, that's correct. The specific replication tests are fixed in this commit: https://github.com/orbitdb/orbit-db/commit/40a0c43ee111fbbca2147ed48a55902abb544a9a

aphelionz:
Merging this and releasing `orbit-db@0.26.0` :tada:

I'll deal with the other stuff later.

PR #853, opened 2020-12-07
fix: dist publish

### LINKED ISSUE #852: should the dist folder be published?
The dist folder is not published (to npm) currently although looking at https://github.com/orbitdb/example-orbitdb-todomvc/ it looks like it is supposed to be.

A script should be added and used to publish that would, at least, build dist and publish.
Unrelated but the [package.files](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#files) field could be configured as well to specify what files/folders should be published, in this future case it would be `["src", "dist"]`.

tabcat:
This implements changes related to publishing mentioned in https://github.com/orbitdb/orbit-db/issues/852

fixes #852

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,10 @@
   "engines": {
     "node": ">=10.0.0"
   },
+  "files": [
+    "src",
+    "dist"
+  ],
   "main": "src/OrbitDB.js",
   "dependencies": {
     "cids": "^1.0.0",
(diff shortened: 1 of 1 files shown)


aphelionz:
We had been avoiding this before, but if this helps you keep TODO MVC up to date then so be it :+1:

PR #854, opened 2020-12-10
Fix/repair examples

tabcat:
Summary of changes
- updated demo 1 link on README
- updated path to ipfs pre-bundle
- updated IPFS api usage to async
- added webrtc signal nodes to swarm config
- fixed up scripts related to the examples

PR #857, opened 2020-12-11
remove obsolete makefile line

tabcat:
this line is no longer needed as a result of https://github.com/orbitdb/orbit-db/pull/854 browser.html not using this folder anymore


aphelionz (approved):
Thanks!

PR #859, opened 2020-12-18
Update README.md

aphelionz:


RichardLitt:
Is Node 8.0.0 now valid, too?

aphelionz:
I'm wondering if we should just get rid of the version numbers entirely and just say LTS. Easier to maintain :D

RichardLitt:
If it works, that works. :)

Issue #867, opened 2021-03-02
Discord invite has expired.


digitalextremist:
The `README.md` invite link to `Discord` has expired, and needs to be updated.

RichardLitt:
The discord was owned by PL, and was either taken over or removed. We currently don't have a discord.

How do we feel about just removing discord everywhere, since it isn't being used and Gitter works for our purposes?

aphelionz:
Yes and now that Gitter speaks matrix we get two for one :)

RichardLitt:
I wasn't happy with these old links existing. There.

phillmac:
Holy pr spam, Batman! 😜

RichardLitt:
Haha I'm sorry. It had to happen, though.

PR #869, opened 2021-03-04
docs: Remove discord link

RichardLitt:
Should I also remove matrix? See #867


tabcat:
definitely keep the matrix link. I just made sure its working. its the same chat room as the one on gitter; discord used to be bridged to there also but that has been phased out. the discord channel is now very inactive and not a good place to send new users to.

RichardLitt:
Cool. I'm hoping to be able to do this for all of the repos automatically using Octoherd: https://github.com/octoherd/octoherd/discussions/15#discussioncomment-479749.

PR #880, opened 2021-04-17
Specified all possible intended behaviors of `DocumentStore.get`

CSDUMMI:
The documentation starts with a statement describing all the possible cases and then describes the edge case of no document matching a key.


aphelionz (approved):
Thank you!

PR #889, opened 2021-05-19
In code documentation with sphinx-js

CSDUMMI:
I have created a documentation directory `docs` where I setup `sphinx` with the `sphinx_js` extension.

Using this we can create `reStructuredText` Files and reference the functions, classes and other values in 
the JS source code of this repository to compile in an easy to read, reference and search static html site.

(Other compilation targets do exist.)

I propose creating further JSDoc documentation (since sphinx-js is based on the syntax of JSDoc) and 
building a CI/CD procedure, that automatically builds and deploys the documentation to html and pins it to IPFS and IPNS. Then using a DNSLink to redirect anyone from `docs.orbitdb.org` to the most recent version of the docs.


haydenyoung:

PR #891, opened 2021-05-21
Add `OrbitDB.Store` property

CSDUMMI:
To make it possible to access the class from the main module,
without the use of another module.

Like the other stores have already been made available.


aphelionz (approved):

CSDUMMI:
Thanks!

PR #912, opened 2021-09-05
Upgrade ipfs and orbit-db deps; use new cid

tabcat:
summary of changes:
- updates path to ipfs dist bundle
- browser test now bundles in development mode
- old cids packages has been removed from dependencies
- [orbit-db-access-controllers moves orbit-db-io to peerDependencies and tests now live here](https://github.com/orbitdb/orbit-db-access-controllers/pull/64)
- [orbit-db-io uses block api](https://github.com/orbitdb/orbit-db-io/pull/26)
- [orbit-db-store updates ipfs-log and moves orbit-db-io to peerDependencies](https://github.com/orbitdb/orbit-db-store/pull/116) and pulls in [replication refactor](https://github.com/orbitdb/orbit-db-store/pull/100)
- orbit-db-test-utils upgraded and orbit-db tests now use ipfs packages from devDependencies
- multiformats/cid now used
- replicate test updated to use ipfs instead of memstore shim


CSDUMMI (approved):
This update is needed in order to work with the current version
of the multiformats/cid package.

tabcat:
some replication tests and the browser tests are failing still. Haven't looked at what is wrong with the replication tests yet. The browser tests are failing locally because of the way they are bundled + https://github.com/ipfs/js-ipfs/issues/3852?notification_referrer_id=MDE4Ok5vdGlmaWNhdGlvblRocmVhZDIzODExMDQ4NDg6MzY5MzMwOTQ%3D#event-5284785075, so they may need to be bundled differently. Not going to fix these until after orbit-db@next is published.

the ci workflow may need to be updated to use node 14 so peerDependencies arent installed  for the orbit-db stores and they are forced to use the new orbit-db-store@next. but this would be fixed by releasing new latest versions of the updated packages instead of using next, so could also be postponed to the full release.

orbit-db-access-controllers currently use orbit-db as a dev dependency to run tests. if possible they should be rewritten to not require orbit-db, but if necessary we should think about moving those tests into orbit-db.

still working on publishing to orbit-db@next with support for the latest ipfs versions. every updated package besides orbit-db-access-controllers could probably just have a new latest published.

CSDUMMI (commented):
I would at least want to merge the changes to this file:
src/orbit-db-address.js

Since users are having problems to use OrbitDB and I think the issue is fixed by this.

tabcat:
unfortunately there are more problems than this for supporting the newer ipfs versions

CSDUMMI:
What other problems are there?

tabcat:
The only reason the tests are failing now is that the package-lock.json file is being deleted before installing and the orbit-db stores are installing their own version of orbit-db-store, due to the orbit-db-store@next version not fitting the peerDependency version requirement.
we need to publish new versions of the updated packages under the `latest` tags next.

aphelionz (approved):
let's do this!

Issue #819, opened 2020-07-13
The Road to 1.0


haadcode:
# The Road to 1.0

by @haadcode and @aphelionz

Big things are coming to OrbitDB. In this document, we describe our proposal for getting OrbitDB from alpha to 1.0.

As always, our roadmap is a combination of the long-term vision for OrbitDB, features the users have been asking for, issues the community wants to address, the core developers have separately discussed, or have been encountered as bugs in using OrbitDB. Please note that inclusion in the road map is not a promise of delivery!

Things here are subject to alteration or deletion. As of this writing, these should be considered proposals and open conversations and anybody should feel welcome and able to provide feedback in the form of questions, comments, or suggestions.

As always, we welcome contributions from the community and would be happy to help to land any of the discussed features or fixes.

If you would like to financially support OrbitDB, we now have an [OpenCollective](https://opencollective.com/orbitdb) that we request you contribute to. Anything helps and we are forever grateful for your support, monetary or otherwise.

In general, the feaqtures and improvements proposed revolve around three categories: Performance and Resource Consumption, User Experience, and Encryption. Without further ado, let's look at the specific items.

## Checklist

- [Non-breaking changes](#non-breaking-changes)
    - [x] [Replicator Refactoring](#replicator-refactoring)
    - [x] [BTree Indexing for KVStore and DocStore](#btree-indexing-for-kvstore-and-docstore)
    - [ ] [Snapshots](#snapshots)
    - [ ] [User Experience Improvements](#user-experience-improvements)
    - [ ] [Developer Experience: The Publish Dance](#developer-experience-the-publish-dance)
    - [ ] [Community efforts](#community-efforts)
- [Breaking Changes](#breaking-changes)
    - [ ] [Oplog Watermarks](#oplog-watermarks)
    - [ ] [Database Encryption](#database-encryption)
    - [ ] [Streaming / Async Iterators](#streaming-async-iterators)
    - [ ] [Hot / Cold Data Separation](#hot-cold-data-separation)
    - [ ] [Misc. Cleanup](#misc-cleanup)
- [ ] [Potential Rust integration](#potential-rust-integration)

## Non-breaking changes

The changes in this section should be able to be implemented without breaking any backwards compatibility or public-facing APIs. Though certain application-level details might change and need to be addressed, by and large these changes should not require a new major version.

### Replicator Refactoring

[...]

PR #914, opened 2021-09-11
Open DB with address without path/name part

CSDUMMI:
This PR is supposed to fix the TODO in the [Road to 1.0](https://github.com/orbitdb/orbit-db/issues/819) Issue, to permit opening a db with an address, that does not have the name/path part.

Related project: [User Experience Improvements](https://github.com/orgs/orbitdb/projects/7#card-44572249)

This PR also changes the tests to expect this new behavior.


tabcat:
> It makes the name part of the orbitdb address optional while still permitting the use of older addresses.

This is great, support for older addresses and backward compatibility is important. At the same time I think its better to not have any new addresses until 1.0 even though you haven't really changed the address at all.

Also I prefer the behavior of erroring if the name exists in the address and is incorrect.

I believe the reasons for this improvement being suggested is how the orbitdb address is used as a unique key to refer to the database like is done [here](https://github.com/orbitdb/orbit-db/pull/914/files#diff-9bc6ea3b3cb771db760ad8d965f8810628b77fbc130234b29b45f974ead68911R438) and simplifying the addresses in general, but I'm not sure. It's probably best to find out the reasons for this suggested change and see if its worth. From looking at the phrasing in the Road to 1.0 issue again it seems like the author is suggesting removing everything from the address but the ipfs CID.
Personally I'm not convinced changing the address format is useful but I'd love to hear why it is. All the reasons I can think of can be handled by using the address.root as the address internally. And I think its valuable for databases to be differentiated by humans somewhat easily.

CSDUMMI:
> I think its valuable for databases to be differentiated by humans somewhat easily.
That's true. But it is also true, that in cases where the OrbitDB Address is only used by
a machine both the `path` and `/orbitdb/` part may be unnecessary.

What is currently missing from OrbitDB (in my opinion) is a function to convert
a manifest CID to OrbitDB Address and vice versa.

haydenyoung:

PR #918, opened 2021-09-30
Added a test for hasCapability in the "grant" section.

CSDUMMI:
Added a test for `hasCapability`created in [PR](https://github.com/orbitdb/orbit-db-access-controllers/pull/67). 

This test will only work, if that PR is merged first and the proper
version of orbit-db-access-controller is published.


aphelionz (changes requested):
1. https://github.com/orbitdb/orbit-db-access-controllers/pull/66 needs to be merged
2. `orbit-db-store` needs to be version bumped (`npm version minor`)
3. `orbit-db`'s package.json file reference to `orbit-db-store` needs to be updated

Issue #924, opened 2021-10-07
CIDs are valid OrbitDB Addresses


CSDUMMI:
If you provide a valid CID to `OrbitDBAddress.isValid` it returns true.

```js
> OrbitDB.isValidAddress("QmTeAppWEwssc4d4soLj1yMeDEZUHtbCzvaGVWKzBYFa25")
true
```

https://github.com/orbitdb/orbit-db/blob/587b2fecaed6cafd735331736c36467955b0293f/src/orbit-db-address.js#L17

This is not the behavior I expected nor is it the behavior I find intuitive.
Not all CIDs are OrbitDBAddresses.

I would like to change this behavior, such that `OrbitDB.isValidAddress` (synonym to `OrbitDBAddress.isValid`) only returns true if the provided value is:
1. An instance of `OrbitDBAddress`
2. Or a string following this pattern: `/orbitdb/<manifest-cid>/<name>`.

Or whatever the valid address format will be in the future.

PR #925, opened 2021-10-07
OrbitDBAddress.isValid and parseAddress accept CIDs.

CSDUMMI:
Refer to this issue for more details: https://github.com/orbitdb/orbit-db/issues/924

`OrbitDB.isValid` returns true and `parseAddress` returns a proper `OrbitDBAddress` object when given a CID as input without name or protocol.

For Example:
```js
> OrbitDB.isValidAddress("QmTeAppWEwssc4d4soLj1yMeDEZUHtbCzvaGVWKzBYFa25")
true
```

This PR will try to add further tests for both of these functions on edge cases and change the behavior of the two functions in order to not accept any Address string that does not conform to this pattern (`<name>` part is optional): `/orbitdb/<manifest-cid>/<name>`

### DIFF
--- a/src/orbit-db-address.js
+++ b/src/orbit-db-address.js
@@ -2,8 +2,6 @@
 const path = require('path')
 const { CID } = require('multiformats/cid')
 
-const notEmpty = e => e !== '' && e !== ' '
-
 class OrbitDBAddress {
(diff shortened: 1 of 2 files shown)


haydenyoung:

PR #928, opened 2021-10-16
Fix replication concurrency

haadcode:
This PR fixes replication concurrency in OrbitDB.

Long ago we made the changes in ipfs-log to support parallel fetching of entries but those updates never made it into OrbitDB side. This PR changes adds those changes thus adding full replication concurrency support to OrbitDB. This in turn means that databases are replicated faster than before in addition to loading faster than before.

This has a major impact on the perceived (and actual) speed of OrbitDB 🎉 

This PR depends on https://github.com/orbitdb/orbit-db-store/pull/117 and https://github.com/orbitdb/ipfs-log/pull/340.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3971,6 +3971,64 @@
       "integrity": "sha512-5tK7EtrZ0N+OLFMthtqOj4fI2Jeb88C4CAZPu25LDVUgXJ0A3Js4PMGqrn0JU1W0Mh1/Z8wZzYPxqUrXeBboCQ==",
       "dev": true
     },
+    "cids": {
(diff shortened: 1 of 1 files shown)


haadcode:
This is still a bit WIP, need to clean up things before ready for review.

Also using a [branch of orbit-db-test-utils](https://github.com/orbitdb/orbit-db-test-utils/tree/fix/paths-to-keys) to make sure tests are running. Need to PR that before this PR can be considered ready.

haadcode:
Why is CI not running?

haadcode:
Will move this to a branch in this repo instead of my fork.

PR #929, opened 2021-10-16
Fix replication concurrency

haadcode:
This PR fixes replication concurrency in OrbitDB.

Long ago we made the changes in ipfs-log to support parallel fetching of entries but those updates never made it into OrbitDB side. This PR adds those changes thus adding full replication concurrency support to OrbitDB. This in turn means that databases are replicated faster than before in addition to loading faster than before.

This has a major impact on the perceived (and actual) speed of OrbitDB 🎉 

This PR depends on https://github.com/orbitdb/orbit-db-store/pull/117 and https://github.com/orbitdb/ipfs-log/pull/340.

### DIFF
--- a/test/replicate.test.js
+++ b/test/replicate.test.js
@@ -22,7 +22,7 @@ const dbPath2 = './orbitdb/tests/replication/2/db2'
 
 Object.keys(testAPIs).forEach(API => {
   describe(`orbit-db - Replication (${API})`, function() {
-    this.timeout(config.timeout)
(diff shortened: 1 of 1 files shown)


haadcode:
This PR also depends on https://github.com/orbitdb/orbit-db-counterstore/pull/44

haadcode:
Tests seem to pass now reliably, beside the random times when ipfs can't find peers. Need to still do a bit of clean up.

PR #936, opened 2021-11-23
require address with prefix

tabcat:
Changes the valid address format to always include the `/orbitdb/` prefix and optionally include the `/<name>` suffix. Previous tests showed addresses starting with the manifest hash as valid.
This might look like a large change on the address format but should have very little impact.

Also cleaned up some parts and added a few tests. Would like to see about moving away from the `path` module if we only need it for the join method.
Another thing, it seems like it would be fine to use posix paths always. What is the reason Windows paths are used sometimes?

### DIFF
--- a/src/orbit-db-address.js
+++ b/src/orbit-db-address.js
@@ -1,8 +1,8 @@
 'use strict'
 const path = require('path')
 const { CID } = require('multiformats/cid')
-
-const notEmpty = e => e !== '' && e !== ' '
+const replaceDelimiter = (address) => address.toString().replace(/\\/g, '/')
+const prefix = '/orbitdb/'
 
 class OrbitDBAddress {
   constructor (root, path) {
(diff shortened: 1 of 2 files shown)


src/orbit-db-address.js
```diff
@@ -15,51 +15,32 @@ class OrbitDBAddress {
   }
 
   static isValid (address) {
```
aphelionz, line 17: I believe this function is at the top of the execution stack and can fail in two ways, as per the tests

1. If the CID is invalid
2. If the address is missing a prefix.

Instead of returning `false` here and the throwing the error in `parse`, what about throwing specific errors here. That would make things more explicit i.e.

1. "Invalid OrbitDB address: xxxx is not a valid CID"
2. "Invalid OrbitDB address: missing /orbitdb/ prefix"

aphelionz (changes requested):
Looks good, just one suggestion

src/orbit-db-address.js
tabcat: I dont think isValid should result in a boolean or an error. Having these two, more specific, errors is a good idea though.

This api is going to be changed soon to have similar methods to cids so I think we keep it the same for now and do this with https://github.com/tabcat/SCPs/blob/SCP-11/scps/SCP-11.md#address-class 's asAddress method.

tabcat:
I'm closing in favor of a larger address change which will include manifest/OrbitDB api changes which will be opened today or tomorrow.

PR #937, opened 2021-11-24
new address and manifest classes

tabcat:
This pr will introduce and use the new classes for the Address and Manifest data types, based on https://github.com/orbitdb/SCPs/pull/12 and https://github.com/orbitdb/SCPs/pull/13.

The goal is to simplify and uniform interactions with them internally.

While integrating the classes inside this pr, OrbitDB methods returning instances of stores will be pruned to just `open`, as discussed with @haadcode, which will take a stringified address.

### DIFF
--- a/src/address.js
+++ b/src/address.js
@@ -1,0 +1,64 @@
+'use strict'
+const { CID } = require('multiformats/cid')
+const prefix = '/orbitdb/'
+const parse = function (address) {
+  address = address.toString()
+
+  if (!address.startsWith(prefix)) {
+    throw new Error(`'${prefix}' prefix missing from address: ${address}`)
+  }
+
+  try {
+    return CID.parse(address.split('/')[2])
+  } catch (e) {
+    console.error(e)
+    throw new Error(`failed to parse CID in address: ${address}`)
+  }
+}
+
+class Address {
+  constructor (cid) {
+    this.cid = cid
+  }
+
+  static get prefix () { return prefix }
+
+  static asAddress (address = {}, force = false) {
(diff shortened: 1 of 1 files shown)


src/manifest.js
```diff
@@ -0,0 +1,68 @@
+'use strict'
+const Address = require('./address')
+const { Block } = require('multformats/block')
+const { sha256: hasher } = require('multiformats/hashes/sha2')
+const mhtype = 'sha2-256'
+const codec = require('@ipld/dag-cbor')
```
tabcat, line 6: the idea is to move this to io and still interact with the Block data type through some encode/decode methods but it take care of the hasher, codec, and subsequent writing to ipfs.

src/manifest.js
```diff
@@ -0,0 +1,68 @@
...
+    this._address = Address.asAddress({ cid: this.cid }, true)
+  }
+
+  static async create (orbitdb, { AccessControllers, name, type, ...options } = {}) {
+    const value = { name, type }
+    const accessControllerAddress = await AccessControllers.create(
+      orbitdb,
+      options.accessController?.type || 'ipfs',
+      options.accessController || {}
+    )
```
tabcat, line 26: it may be best to have the accessControllers use the `Manifest.create` method to create their manifest eventually.
As part of 1.0, I'd like to see the orbit-db accessController to not use the ipfs accessController, creating another manifest within a manifest within a manifest. This would make it easier to tell that: if we have the manifest and the accessController options resolved, then we have everything we need to open and run the store.

PR #944, opened 2021-12-30
Documentation of the ordering in oplog.values

CSDUMMI:
The order of `oplog.values` is not explicitly defined.

### DIFF
--- a/API.md
+++ b/API.md
@@ -674,7 +674,7 @@ Oplog or IPFS-Log is the underlying data structure used by the OrbitDB Stores to
 replicate and add operations to any of their stores. 
 
 ## `oplog.values`
-An Array of operations sorted in ascending order by their time.
+An Array of operations sorted by how they happened in ascending order.
 If an operation A is before B, A is before B in the Array.
 
 Each oplog entry contains these fields:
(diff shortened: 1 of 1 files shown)


tabcat:
would you mind linking to the source just so i can look at the current order rules

tabcat:
sorry i read the pr title wrong, this seems fine and doesnt have to go into ordering. :+1:

API.md
```diff
@@ -674,6 +674,9 @@ Oplog or IPFS-Log is the underlying data structure used by the OrbitDB Stores to
 replicate and add operations to any of their stores. 
 
 ## `oplog.values`
+An Array of operations sorted in ascending order by their time.
+If an operation A is before B, A is before B in the Array.
```
tabcat, line 678: I'd try to leave the term 'time' out of this

CSDUMMI: What is it ordered by if not time? Causal relationship?

tabcat: yes and a logical clock but its not necessary to go into that here.

tabcat (approved):

API.md
```diff
@@ -674,6 +674,9 @@ Oplog or IPFS-Log is the underlying data structure used by the OrbitDB Stores to
 replicate and add operations to any of their stores. 
 
 ## `oplog.values`
+An Array of operations sorted by how they happened in ascending order.
+If an operation A is before B, A is before B in the Array.
```
tabcat, line 678: i probably would have used entries here instead of operations but it might actually be an array of operations depending on how batched operations are implemented so its fine

CSDUMMI:
Thanks!

Issue #952, opened 2022-01-25
IdentityProvider type 'ethereum' is not supported


anarkrypto:
The project [orbit-db-identity-provide](https://github.com/orbitdb/orbit-db-identity-provider#creating-an-identity-with-an-ethereum-wallet)  supports **creating an identity with an Ethereum wallet**.

Following the step-by-step, I can connect to my Metamask provider and sign Identity. However, if I try to connect to Orbit-DB using this identity, this is the error I get in Orbit-DB.

```
2022-01-25T08:08:11.186Z [ERROR] orbit-db: Error: IdentityProvider type 'ethereum' is not supported
    at getHandlerFor (/root/services/orbitDB/node_modules/orbit-db-identity-provider/src/identities.js:19:11)
    at Identities.verifyIdentity (/root/services/orbitDB/node_modules/orbit-db-identity-provider/src/identities.js:92:30)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async saveToIpfs (/root/services/orbitDB/node_modules/orbit-db-store/src/Store.js:317:25)
    at async pEachSeries (/root/services/orbitDB/node_modules/p-each-series/index.js:8:23)
    at async OrbitDB._onMessage (/root/services/orbitDB/node_modules/orbit-db/src/OrbitDB.js:263:11)
```

orbit-db: 0.28.1
orbit-db-identity-provider: 0.4.0

Implemented with Node.JS

tabcat:
My guess is that the orbit-db-identity-provider dependency in orbitdb needs to be updated to version 4

https://github.com/orbitdb/orbit-db/blob/726e667a4f452876271a3a40d22999268cc8a3e2/package.json#L32

anarkrypto:
@tabcat 

Updating the version of **orbit-db-identity-provider** to **^0.4.0** in **orbit-db/package.json** fixed it.
Thank you very much!

I think we'd better make a pull request now, don't you?

haydenyoung:
Thanks for the PR. Closing.

PR #953, opened 2022-01-26
update identity-provider-interface to ^v0.4.0

anarkrypto:
Solve the issue **IdentityProvider type 'ethereum' is not supported**  #952


tabcat:
Looks like we may need to add a rule for the did-jwt module to be loaded with webpack.

anarkrypto:
@tabcat  Any update ?

anarkrypto:
orbit-db v29.0 is already implementing identity-provider-interface v0.5.

Issue #911, opened 2021-09-02
switch puppeteer for puppeteer-core


tabcat:
this saves some time by not automatically installing the puppeteer chromium binaries; we can choose to install them before running browser tests.

below are the repos in the org that are affected
- https://github.com/orbitdb/orbit-db

repos using puppeteer via `mocha-headless-chrome` should also be changed to using puppeteer-core but would be a little more difficult
- https://github.com/orbitdb/ipfs-log/blob/main/package.json
- https://github.com/orbitdb/orbit-db-store/blob/main/package.json
- https://github.com/orbitdb/orbit-db-storage-adapter/blob/main/package.json
- https://github.com/orbitdb/orbit-db-keystore/blob/main/package.json
- https://github.com/orbitdb/orbit-db-cache/blob/main/package.json
- https://github.com/orbitdb/orbit-db-identity-provider/blob/main/package.json

the goal is to only download the chromium binaries if the developer is running browser tests so better ways to solve this may exist. the optimal solution is one that could be used directly from an npm script (puppeteer and puppeteer-core cannot; mocha-headless-chrome can) and waits to download the chromium binaries (puppeteer-core allows this via the install script).

PR #965, opened 2022-02-23
fix(#911): Swithing puppeteer to puppeteer core

captaincole:
Fixes this issue: https://github.com/orbitdb/orbit-db/issues/911

Specifically we:
* Puppeteer to Puppeteer Core
* Because of an issue with M1 Macs Chromium problems, we include chromium as a dev dependency [See this issue](https://github.com/puppeteer/puppeteer/issues/6641)


tabcat:
Thanks for making this, my original intention moving it to puppeteer-core was to delay chromium binary installation until the tests were being ran the first time. Since it looks like pupperteer is broken for M1 macs and the chromium package can also [delay installation](https://github.com/dtolstyi/node-chromium#skip-automatic-chromium-install) this looks great. We can add the delay install part later.

Thanks again for making this pr @captaincole !
I'll be sure to check it out closer and merge it today.

PR #986, opened 2022-03-30
Add .nvmrc and specifying support for Node 16

### LINKED ISSUE #985: Doc Node Version support out of date. NVM potential?
Hello! I hope everything has been going well,
Node Version: `v14.17.4`

When I was initially trying to set the code up to play around with, I noticed that I was receiving a weird error telling me
> class TimeoutController extends AbortController {
TypeError: Class extends value undefined is not a constructor or null

I checkout out the Element lobby and someone suggested upping your node version from v14 to v16. 
[I noticed the docs saying that it supports all even numbered node versions](https://github.com/orbitdb/orbit-db/blob/main/README.md?plain=1#L31)  but I didn't find that to be the case for my v14 node 😔

I was wondering if maybe the docs could be updated to specify that people should be using a node version >16?
[To kind of extend on that idea, adding a `.nvmrc` so users could `nvm use`](https://github.com/nvm-sh/nvm) and get the right node version might be helpful? Any thoughts?

caleb-mabry:
Resolves #985 

Adding support for `nvm` by including a `.nvmrc` and specifying that orbit-db works with Node 16.


.nvmrc
```diff
@@ -0,0 +1 @@
+14.17.4
```
tabcat, line 1: guessing this will be changed to 16

caleb-mabry: 🤦 
You're right

tabcat:
@caleb-mabry do you think you could look at the [node](https://www.npmjs.com/package/orbit-db) badge in the readme? i couldnt quite figure it out.

PR #991, opened 2022-04-24
Make 'keystore' not undefined

aspironical:
This PR fixes `TypeError: Cannot read properties of undefined (reading 'keystore')` when `options.identity` is defined and `options.identity.provider.keystore` is not.


tabcat:
Hi @aspironical 

When is this the case? My thought is that if there is an identity that doesn't fit this convention it, then it should be changed so it does. I thought all Identity instances had this property.

PR #992, opened 2022-04-24
Fix fs-shim error in electron renderer process

julienmalard:
is-electron returns true for the electron renderer process, which leads to a `require is not defined` error. Using wherearewe instead and explicitly checking for the electron main process resolves this issue.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,6 @@
       "license": "MIT",
       "dependencies": {
         "ipfs-pubsub-1on1": "0.0.8",
-        "is-electron": "^2.2.0",
         "is-node": "^1.0.2",
         "localstorage-down": "^0.6.7",
         "logplease": "^1.2.14",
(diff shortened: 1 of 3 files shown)


tabcat:
This new env detection module looks great.

Thanks for the description of the issue. Will take a closer look and probably merge and publish.

Any ideas how to do this write for ESM, preferably by static import. Does package.browser cover electron?

julienmalard:
Thansk! I'm just starting to get comfortable with esm modules, but I'm pretty sure that package.browser would cover the electron renderer process because it gets compiled for the browser when electron is packaged.

Could we just use `import fs from "fs"` use a polyfill when compiling for the browser?

tabcat:
I don't think polyfilling fs for browser envs will necessary unless its just polyfilled as undefined. we will likely be using level for most stuff, may need to make a few directories when in a node env but that will be it for fs writes.

PR #1007, opened 2022-06-24
wrong ipfs.js reference in browser examples 

inspiraluna:
1.  corrected reference to ipfs.js
2.  corrected build instructions so browser libs are correctly copied
3.  added workaround to README in case browser isn't opening (under MacOS)

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,14 +160,14 @@ to redo the local package-lock.json with working native dependencies.
 ### Browser example
 
 ```
-npm run build
-npm run examples:browser
+make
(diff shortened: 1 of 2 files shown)


tabcat:
Looks good. Is `open-cli` not working for mac? Maybe it needs to be updated?

inspiraluna:
I checked into that too a bit. There seems to be an issue with open-cli on Mac. https://github.com/sindresorhus/open-cli/issues/12

tabcat:
Will merge and make some changes so testing is not ran.

PR #1010, opened 2022-07-10
Fix for "Ipfs not defined in webpack example"

silkroadnomad:
When running the examples as described in Readme, after opening the browser console shows "Ipfs not found".
The import Ipfs from "ipfs" doesn't seem to resolve for some reason. But if replaced with a

const Ipfs = require('ipfs')

the whole example works like a sharm (as the other browser example too)
We fixed a similar problem in the other browser example lately but didn't see that the webpack example doesn't work either. Now it does.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ test: deps
 	npm run test:all -- --exit
 	npx validate-maintainers orbit-db --match --ci
 
-build: test
+build:
 	mkdir -p examples/browser/lib/
 	npm run build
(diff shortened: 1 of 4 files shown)


tabcat:
Would be good to know when this broke, I guess this is a case for more tests :grin: 

Will check this out more and come back. Thanks for reporting this.

tabcat:
Guessing this has to do with the package resolution for ipfs changing? Will merge now after confirming fix, but would be nice to know exactly what is happening here.

tabcat:
have to fix something with the Makefile and then the checks should pass.

PR #1013, opened 2022-08-18
Small: cleaned up code in disconnect

CSDUMMI:
Two improvements to the style of the OrbitDB.disconnect function:
- it uses both forEach and for loops to close connections. I removed the one instance of using forEach and replaced it by using a for loop.
- it repeatedly deletes every element in an object using `delete`. I removed these and reassigned the variable to an empty object at the end of the loop. I believe this would be more efficient, because garbage collectors do not have to reallocate the object piece meal.

### DIFF
--- a/src/OrbitDB.js
+++ b/src/OrbitDB.js
@@ -166,15 +166,17 @@ class OrbitDB {
     return this.docs(address, options)
   }
 
+  /**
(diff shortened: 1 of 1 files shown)


aphelionz (approved):
I like it!

PR #1090, opened 2023-09-03
refactor: Identity provider can be passed as partially instantiated f…

haydenyoung:
…unction.


src/orbitdb.js
```diff
@@ -22,13 +22,16 @@ const DefaultAccessController = IPFSAccessController
...
+ * @param {String} [params.identity.id] An initialized identity provider.
+ * @param {Function} [params.identity.provider] An initialized identity provider.
  * @param {module:Identities} [params.identities] An Identities system instance.
  * @param {string} [params.directory] A location for storing OrbitDB data.
  * @return {module:OrbitDB~OrbitDB} An instance of OrbitDB.
  * @throws "IPFS instance is required argument" if no IPFS instance is provided.
  * @instance
  */
-const OrbitDB = async ({ ipfs, id, identities, directory } = {}) => {
+const OrbitDB = async ({ ipfs, id, identity, identityProvider, identities, directory } = {}) => {
```
haadcode, line 34: The `identityProvider` is not needed anymore I believe.

src/orbitdb.js
```diff
@@ -51,7 +54,13 @@ const OrbitDB = async ({ ipfs, id, identities, directory } = {}) => {
     identities = await Identities({ ipfs, keystore })
   }
 
-  const identity = await identities.createIdentity({ id })
+  if (identity) {
+    if (identity.provider) {
+      identity = await identities.createIdentity({ id, provider: identity.provider })
```
haadcode, line 59: Should the `id` here be `id: identity.id`?

src/identities/providers/publickey.js
```diff
@@ -27,7 +27,7 @@ const verifyIdentity = identity => {
  * identity provider function.
  * @private
  */
-const PublicKeyIdentityProvider = ({ keystore }) => {
+const PublicKeyIdentityProvider = ({ keystore }) => () => {
```
haadcode, line 30: Wondering if we should make the second resented function async? That’s how stores and ACs are and may be useful for 3rd party devs to do async io inside the closure.

haadcode: If so, need to add `await` to https://github.com/orbitdb/orbitdb/pull/1090/files#diff-8c445a0a80e9ec75a67dd06dcd352e8bfac45b3a9c67d662a65400c253141733R76

src/databases/keyvalue-indexed.js
```diff
@@ -21,6 +21,7 @@ import KeyValue from './keyvalue.js'
 import LevelStorage from '../storage/level.js'
 import pathJoin from '../utils/path-join.js'
 
+const type = 'keyvalueindexed'
```
haadcode, line 24: This should be the same as KeyValue.type as we want keyvalue-indexed to handle all keyvalue databases.

src/databases/index.js
```diff
@@ -55,4 +56,9 @@ const getDatabaseType = (type) => {
   return databaseTypes[type]
 }
 
+addDatabaseType(Events)
+addDatabaseType(Documents)
+addDatabaseType(KeyValue)
+addDatabaseType(KeyValueIndexed)
+
```
haadcode, line 63: Noticing this only now but there prolly should be `removeDatabaseType` also so that the user can remove a default handler and add their custom handler for any database type.

src/identities/providers/index.js
```diff
@@ -18,29 +16,30 @@ const getIdentityProvider = (type) => {
...
+  if (identityProviders[identityProvider.type]) {
+    throw new Error(`IdentityProvider ${identityProvider.type} already added.`)
   }
 
-  identityProviders[IdentityProvider.type] = IdentityProvider
+  identityProviders[identityProvider.type] = identityProvider
 }
 
+addIdentityProvider(PublicKeyIdentityProvider)
+
```
haadcode, line 44: We prolly should have `removeIdentityProvider` here to make it consistent with the others?

src/databases/keyvalue-indexed.js
haydenyoung: So does this mean we shouldn't register the [keyvalue-indexed database here](https://github.com/orbitdb/orbitdb/blob/a0fc91d67497723600c80742042c0608c7d17ddb/src/databases/index.js#L62) because the implementer should be able to remove the keyvalue database and replace with the keyvalue-indexed?

haadcode: Correct. The registration effects which database orbitdb uses by default when opening a db (or creating one). If the user wants to open a db of type `keyvalue` they can open it without registering the database type and passing it as a param to the open() fn. If KVIndexed is registered, then it’ll be used (instead of KeyValue) as the default when opening a db of type `keyvalue`.

src/databases/index.js
haydenyoung: Removing removeX and using useX (as discussed).

docs/DATABASES.md
```diff
@@ -244,13 +246,29 @@ const CustomDB = async ({ OpLog, Database, ipfs, identity, address, name, access
...
+export default CustomDB
 ```
 
 [Documents](../src/db/documents.js), [Events](../src/db/events.js) and [KeyValue](../src/db/keyvalue.js) provide good examples of how a database is implemented in OrbitDB and how to add the logic for returning records from the database (the state of the database).
+
+To use a custom database, add it to the list of supported database types:
+
+```js
+import { createOrbitDB, useDatabaseType } from '@orbitdb/core'
+import { CustomDB } from './custom-db.js'
```
haadcode, line 268: Should this be `import CustomDB from…`?

docs/IDENTITIES.md
```diff
@@ -143,7 +143,7 @@ const verifyIdentity = identity => {
 }
 
 // The identity provider.
-const MyCustomIdentityProvider = ({ keystore }) => {
+const MyCustomIdentityProvider = ({ keystore }) => () => {
```
haadcode, line 146: Should add `async` here for the returned function to indicate that dev can do async io within the IP.

docs/IDENTITIES.md
```diff
@@ -168,7 +172,7 @@ To use it, add it to the list of known identity providers:
 import { addIdentityProvider } from '@orbitdb/core'
 import MyCustomIdentityProvider from 'my-custom-identity-provider'
 addIdentityProvider(MyCustomIdentityProvider)
-const identity = await createIdentity({ id: 'some id', type: 'custom' })
+const identity = await createIdentity({ type: MyCustomIdentityProvider(options) })
```
haadcode, line 175: Here `type` should be `provider` (https://github.com/orbitdb/orbitdb/pull/1090/files#diff-8c445a0a80e9ec75a67dd06dcd352e8bfac45b3a9c67d662a65400c253141733R67)

src/identities/identities.js
```diff
@@ -66,17 +64,21 @@ const Identities = async ({ keystore, path, storage, ipfs } = {}) => {
...
    * @memberof module:Identities~Identities
    * @instance
    */
   const createIdentity = async (options = {}) => {
     options.keystore = keystore
+    const DefaultIdentityProviderType = getIdentityProvider('publickey')
+    const IdentityProvider = options.provider || DefaultIdentityProviderType({ keystore })
+    const identityProvider = await IdentityProvider()
+
+    if (!getIdentityProvider(identityProvider.type)) {
```
haadcode, line 78: We can do this check now with `IdentityProvider.type` before L76

test/fixtures/providers.js
```diff
@@ -1,7 +1,7 @@
 const customIdentityProvider = () => {
   const verifyIdentity = async (data) => { return true }
 
-  const CustomIdentityProvider = () => {
+  const CustomIdentityProvider = () => () => {
```
haadcode, line 4: Add `async` to the returned function?

test/fixtures/providers.js
```diff
@@ -23,7 +23,7 @@ const customIdentityProvider = () => {
 const fakeIdentityProvider = () => {
   const verifyIdentity = async (data) => { return false }
 
-  const FakeIdentityProvider = () => {
+  const FakeIdentityProvider = () => () => {
```
haadcode, line 26: Add `async` to the returned function?

test/fixtures/providers/custom.js
```diff
@@ -0,0 +1,20 @@
+const type = 'custom'
+
+const verifyIdentity = async (data) => { return true }
+
```
haadcode, line 4: Add `async`?

test/fixtures/providers/fake.js
```diff
@@ -0,0 +1,20 @@
+const type = 'fake'
+
+const verifyIdentity = async (data) => { return false }
+
+const FakeIdentityProvider = () => () => {
```
haadcode, line 5: Add `async`?

src/access-controllers/index.js
```diff
@@ -57,9 +48,12 @@ const removeAccessController = type => {
...
 }
 
+useAccessController(IPFSAccessController)
+useAccessController(OrbitDBAccessController)
+
 export {
   getAccessController,
-  addAccessController,
+  useAccessController,
   removeAccessController,
```
haadcode, line 57: Remove

haadcode (commented):
Looking great! 💪 Couple of small suggestions.

src/identities/identities.js
haydenyoung: Wasn't able to implement this:

1. IdentityProvider.type <- this exists
2. ~~IdentityProviderPartial.type~~ <- Partial (I.e. `IdentityProvider(options)`  is passed in as the `provider` param and doesn't have a type
3. identityProviderInstance.type <- this exists

When testing whether the IdentityProvider is being used, 2 does not expose a type so using 3 is the only option to retrieve the type.

haadcode: Ah, ok. Then let’s leave it as is 👍

haadcode:
LGTM 👍

PR #1095, opened 2023-09-10
Chore/change arrow functions

bitcoinbrisbane:
Some of the tests cases seem to use `function` notation, where others use arrow functions.  This PR addresses that.


haydenyoung:
Unit tests are failing with:

```
TypeError: Cannot read properties of undefined (reading 'timeout')
```

Looks like the `this` for any `this.timeout()` is undefined.

Also noticed CI isn't running on this PR. Not sure why.

haydenyoung:
I'm going to suggest we go with [Mocha's recommendation](https://mochajs.org/#arrow-functions) on this one. I'd prefer not to introduce complexity for the sake of brevity.

Thoughts?

bitcoinbrisbane:
> I'm going to suggest we go with [Mocha's recommendation](https://mochajs.org/#arrow-functions) on this one. I'd prefer not to introduce complexity for the sake of brevity.
> 
> Thoughts?

Ok, as mentioned on the PR the code base has inconsistent styles.  My suggestion would be to make it uniform.  Also feels like PRs are not welcomed on this project.

PR #1099, opened 2023-09-13
Close databases when orbitdb is stopped

haadcode:
This PR fixes a bug where calling `orbitdb.stop()` doesn't close the open databases and upon restarting orbitdb, trying to open a database that was previously open throws an error.

### DIFF
--- a/test/orbitdb-open.test.js
+++ b/test/orbitdb-open.test.js
@@ -529,15 +529,18 @@ describe('Open databases', function () {
   })
 
   describe('opening same database', () => {
-    let db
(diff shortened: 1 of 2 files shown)


test/orbitdb-open.test.js
```diff
@@ -529,15 +529,18 @@ describe('Open databases', function () {
   })
 
   describe('opening same database', () => {
-    let db
+    let db1, db2
```
haadcode, line 532: The changes to this test are merely a clean up.

PR #1108, opened 2023-09-23
Integrate Helia

haydenyoung:
WIP. DO NOT MERGE.

All node tests (`npm run test`) are working.

Some confirmations around the sync protocol:

### Checking for topic in `handleUpdateMessage`

I'm assuming [`handleUpdateMessage`](https://github.com/orbitdb/orbitdb/blob/31597d708420703bc921da02662c594ead446570/src/sync.js#L217) is fired on every message sent by a publishing node, not just the message with the specific topic. Therefore, should we be checking for the topic? E.g.
```
if (msg.topic === address) {
  const task = async () => {
    ...
  }
} 
```

### Decoding data when checking message in `handleUpdateMessage`

Would we need to decode `message.detail.data` to [check whether it is undefined](https://github.com/orbitdb/orbitdb/blob/31597d708420703bc921da02662c594ead446570/src/sync.js#L221)?


haydenyoung:
> Checking for topic in handleUpdateMessage

It appears the old message.data is also a byte array, so this doesn't need to be addressed.

haydenyoung:
https://github.com/orbitdb/orbitdb/blob/31597d708420703bc921da02662c594ead446570/src/sync.js#L220

Do we need to check whether message is from self? pubsub module has a param called `emitSelf` which determines whether the publishing peer receives its own message; it is false by default.

justin0mcateer:
We encountered a bug in the original Helai patch set that was caused by not providing a timeout to Helia. If there is any problem reading or writing to the underlying transport stream, it will hang forever.

haadcode:
> https://github.com/orbitdb/orbitdb/blob/31597d708420703bc921da02662c594ead446570/src/sync.js#L220
> 
> Do we need to check whether message is from self? pubsub module has a param called `emitSelf` which determines whether the publishing peer receives its own message; it is false by default.

Can prolly remove it?

haydenyoung:
All tests are now passing but changing some of the Libp2p settings was required. In particular, disconnectThreshold, which defaults to "5", has been set to "128" to ensure all tests pass. This is probably an unrealistic setting and will likely break if OrbitDB syncs 100s or 1000s of records. Instead, we should determine the cause of the default disconnectThreshold being insufficient and update the sync protocol accordingly.

The following error message may be a clue to why Libp2p is prematurely closing connections unless the disconnectThreshold is excessively high:

```
libp2p:mplex rate limit hit when receiving messages for streams that do not exist - closing remote connection +0ms
```

I'm not sure exactly what is causing this error except that it appears to be related to some kind of DoS protection. My understanding is that the multiplexer, mplex, is being deprecated so we should probably look to implement yamux. As yamux appears to have better flow control mechanisms which might align with the error above, I.e. that the stream could be sending data then closing before the peer can receive all the data and process it (just speculating and don't know enough about this to give more than a high-level opinion).

haydenyoung:
> All tests are now passing but changing some of the Libp2p settings was required. In particular, disconnectThreshold, which defaults to "5", has been set to "128" to ensure all tests pass. This is probably an unrealistic setting and will likely break if OrbitDB syncs 100s or 1000s of records. Instead, we should determine the cause of the default disconnectThreshold being insufficient and update the sync protocol accordingly.
> 
> The following error message may be a clue to why Libp2p is prematurely closing connections unless the disconnectThreshold is excessively high:
> 
> ```
> libp2p:mplex rate limit hit when receiving messages for streams that do not exist - closing remote connection +0ms
> ```
> 
> I'm not sure exactly what is causing this error except that it appears to be related to some kind of DoS protection. My understanding is that the multiplexer, mplex, is being deprecated so we should probably look to implement yamux. As yamux appears to have better flow control mechanisms which might align with the error above, I.e. that the stream could be sending data then closing before the peer can receive all the data and process it (just speculating and don't know enough about this to give more than a high-level opinion), it is probably worth the effort to upgrade.

Looks like changing to yamux fixes this. I'm guessing Yamux doesn't suffer from the same problem because the flow of data is controlled so one can't "flood" the other peer. I'm going to push a yamux upgrade to this branch but will also get feedback from the libp2p devs on the exact reason why Yamux doesn't experience the same issues as Mplex.

haydenyoung:
Response from @achingbrain on above confirms the Mplex/Yamux assumptions above:

> Yes - mplex has no concept of backpressure so the only thing it can do is to close the incoming connection and is deprecated for this reason.
> 
> Yamux is a bit more sane and can throttle remote peers by lowering the amount of data they are allowed to send which slows them down.

src/storage/ipfs-block.js
```diff
@@ -38,14 +38,11 @@ const IPFSBlockStorage = async ({ ipfs, timeout, pin } = {}) => {
...
     const cid = CID.parse(hash, base58btc)
-    await ipfs.block.put(data, {
-      cid: cid.bytes,
-      version: cid.version,
-      format: 'dag-cbor',
-      mhtype: 'sha2-256',
-      pin,
-      timeout
-    })
+    await ipfs.blockstore.put(cid, data)
```
haadcode, line 41: Do we have a way to control the timeout time anymore? If not, we could remove the `timeout` option from the `IPFSBlockStorage` function above?

haydenyoung: I'm not seeing it documented anywhere so probably no longer applies.

Can probably remove:

https://github.com/orbitdb/orbitdb/blob/0a5f8b54e3dac37ebf0ef9446809323588ac75d6/src/storage/ipfs-block.js#L10

and 

https://github.com/orbitdb/orbitdb/blob/0a5f8b54e3dac37ebf0ef9446809323588ac75d6/src/storage/ipfs-block.js#L29

and remove the timeout param from https://github.com/orbitdb/orbitdb/blob/0a5f8b54e3dac37ebf0ef9446809323588ac75d6/src/storage/ipfs-block.js#L26C41-L26C48.

PR #1113, opened 2023-10-18
missing async caused confusion

silkroadnomad:
When copy & pasting the example code it never worked the first time. Now it does.


haadcode:
Good catch! Thank you 🙏

PR #1129, opened 2023-11-20
Release/v1.0.1

haydenyoung:


aphelionz (approved):
Assuming the `v1.0.0` -> `v1.0` change is intentional (as opposed to `v1.0.1`)

haydenyoung:
> Assuming the `v1.0.0` -> `v1.0` change is intentional (as opposed to `v1.0.1`)

Correct. We decided that API docs will not change between patch releases and therefore don't warrant an iteration.

haadcode (approved):

PR #1132, opened 2023-11-30
Improve Oplog's joinEntry implementation

### LINKED ISSUE #1110: joinEntry seems to traverse entire log in certain cases
In the event that we have been disconnected and there are concurrent updates, OpLog.joinEntry seems to cause OpLog.traverse to traverse the _entire_ log to 'double check' that the entry isn't already in the log.

Perhaps I am mis-understanding, but the 'shouldStopFn' value seems that it will never match and thus the traverse will continue to the end of the log. 

In the event that the new entry (Head) points _directly_ to a log entry that already exists locally, this is stop very quickly and function as expected. This seems to correlate with the behavior that we see where updates (even thousands) are handled quite well, so long as all of the nodes are always connected when updates are happening. However, if a node is disconnected and there are updates, the node will hang traversing the entire previous log. 

https://github.com/orbitdb/orbitdb/blob/main/src/oplog/log.js#L238

Why is this extra check applied? Is there a known condition where we might receive an entry which is in the log, but somehow not in the index, or is this just paranoia?

haadcode:
This PR introduces a new implementation of the Oplog's `joinEntry` function.

The implementation changes the way entries are joined by 1) traversing the missing entries between the entry-to-be-joined and current heads or entries in the log and 2) verifies each new entry as they are traversed. This is more efficient than the previous implementation and adds to the security by making sure all entries are verified before adding them to the (local) oplog.

Added a bunch of tests to cover the various forking/joining cases, exposed `Entry.decode()` for testing purposes and added `Heads.remove()` to remove heads one by one used in the new implementation.

Fixes #1110.

### DIFF
--- a/src/oplog/log.js
+++ b/src/oplog/log.js
@@ -128,7 +128,6 @@ const Log = async (identity, { logId, logHeads, access, entryStorage, headsStora
     const bytes = await _entries.get(hash)
     if (bytes) {
       const entry = await Entry.decode(bytes)
-      await _index.put(hash, true)
       return entry
     }
   }
@@ -206,12 +205,12 @@ const Log = async (identity, { logId, logHeads, access, entryStorage, headsStora
     if (!isLog(log)) {
       throw new Error('Given argument is not an instance of Log')
     }
+    if (_entries.merge) {
+      await _entries.merge(log.storage)
+    }
     const heads = await log.heads()
     for (const entry of heads) {
       await joinEntry(entry)
-    }
-    if (_entries.merge) {
-      await _entries.merge(log.storage)
     }
   }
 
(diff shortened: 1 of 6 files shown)


haadcode:
> Does this mean it's no longer neccessary to traverse/iterate the log to achieve log entry replication?

Correct 👍

src/oplog/log.js
```diff
@@ -222,54 +221,86 @@ const Log = async (identity, { logId, logHeads, access, entryStorage, headsStora
...
+    const connectedHeads = new Set()
+
+    const traverseAndVerify = async () => {
+      const getEntries = Array.from(hashesToGet.values()).filter(has).map(get)
+      const entries = await Promise.all(getEntries)
+
+      for (const e of entries) {
+        hashesToGet.delete(e.hash)
+
+        await verifyEntry(e)
```
leblowl, line 269: I wonder if there is a case where you'd like to skip a single invalid entry? It looks like an exception in `verifyEntry` would cause `joinEntry` to stop. But maybe there are cases where you'd like to continue with other `nexts`.

leblowl (commented):
Looks good! I know this has been merged, but I added one thought

src/oplog/log.js
haadcode: There should not be any invalid entries in any part of the oplog DAG. If there were, it’d be a malformed DAG (malicious or otherwise).

PR #1156, opened 2024-01-26
docs: Audit GETTING STARTED examples. Ensure they are running.

haydenyoung:


haydenyoung:
@haadcode as a docs update I think I'm happy with this now and am ready to move it out of WIP and get a review.

docs/GETTING_STARTED.md
```diff
@@ -287,7 +284,13 @@ Copy the database's address from terminal 1 and, in terminal 2, run:
...
 ```
 
-Upon connection, you should see the records being created in terminal 1's database received by the database running in terminal 2.
+Both peers will print new records to the terminal as the log is updated. When you stop each peer using ctrl+c, the final state of the database will be printed to the terminal. They should match.
+
+**PLEASE NOTE:** 
+
+This example is using mDNS to find peers on a local network. This example will not work if each peer is on a different network and you will need to implement an alternative peer discovery mechanism to achieve connectivity. Alternatively, if the address of one of the peers is known and is accessible, the other peer can dial it manually. 
+
+These kinds of connectivity configurations are beyond the scope of OrbitDB.
```
haadcode, line 293: Might we add something regarding where the user can find more info on connectivity conf? “refer to libp2p documentation” or smth?

haadcode (approved):
Couple of suggestions, otherwise looks good to merge 👍

PR #1193, opened 2024-08-23
await verifyIdentity because it's async

leblowl:


haydenyoung:
Merged. Thanks for your contribution.

PR #1196, opened 2024-09-24
Update IDENTITIES.md

Jon-Biz:
Helia has a `blockstore`, rather than a `block` so this example wouldn't run.

https://helia.io/interfaces/helia.Helia.html#blockstore


haydenyoung:
Thanks for your contribution.

Issue #1163, opened 2024-02-22
Log Entries created on remote/other Peers are not Pinned Locally


justin0mcateer:
In the previous implementation of OpLog.joinEntry, Log entries retrieved from other nodes were 'put' in the IPFSBlockStorage and optionally (via configuration) pinned. In the rewrite of OpLog.joinEntry _entries.put is no longer called and so the remotely created Log entries are no longer stored locally.

This results in the behavior that _all_ Peers which have previously written to the log MUST be online or traversal of the Log will fail and unhandled exception in IPFSBlockStorage.get. This does not seem to be desirable behavior.

This code will crash with an unhandled promise rejection if any block fails to be fetched, which currently appears to occur if any previous log writer is offline:
https://github.com/orbitdb/orbitdb/blob/main/src/storage/ipfs-block.js#L61

In older versions of the code (prior to the OpLog.joinEntry rewrite), fetched OpLog entries were persisted locally:
https://github.com/orbitdb/orbitdb/blob/v1.0.0/src/oplog/log.js#L269

haydenyoung:
Is the peer using an in-memory blockstore for Helia? What happens if you configure the node with a persistent blockstore?

E.g.

```
import { LevelBlockstore } from 'blockstore-level' 

const blockstore = new LevelBlockstore('./some/path')
const libp2p = await createLibp2p()
const helia = await createHelia({ libp2p, blockstore })
```

My understanding is that the replicated block should also be pinned on the remote peer (`isPinned` should == "true" on the remote peer also).

haydenyoung:
Just to add that a persistent datastore may also be needed as I believe pins are stored to datastore not blockstore.

```
import { LevelBlockstore } from 'blockstore-level' 
import { LevelDatastore } from 'datastore-level' 

const blockstore = new LevelBlockstore('./some/path/helia/blocks')
const datastore = new LevelDatastore('./some/path/heliadata')
const libp2p = await createLibp2p()
const helia = await createHelia({ libp2p, blockstore, datastore })
```

(sorry haven't been able to actually confirm this, just going on a hunch re: datastore and pins).

justin0mcateer:
We are using a persistent store (IDB) for both. What I am saying is that the new logic doesn't call Helia.put for remote originated blocks at all, at least from what I can see.

There was an explicit call to _entries.put prior to the joinEntry rewrite. I've looked and I can't see it being done anywhere else. We are seeing block not found errors when not all the nodes are online.


[...]

PR #1198, opened 2024-10-15
Helia v5

haydenyoung:


test/orbitdb-replication-pins.test.js
```diff
@@ -0,0 +1,57 @@
...
+
+    const db2 = await orbitdb2.open(db1.address)
+
+    db2.events.on('join', onJoin)
+
+    await waitFor(() => replicated, () => true)
+
+    const cid = CID.parse(hash, base58btc)
+    strictEqual(await ipfs1.pins.isPinned(cid), true)
+    strictEqual(await ipfs2.pins.isPinned(cid), false)
```
haydenyoung, line 55: To pin the cid on the replicating ipfs (ipfs2), the cid needs to be manually pinned, I.e.

```
await drain(ipfs2.pin.add(cid))
```

This will result in ipfs2.pins.isPinned(cid) being true.

cristianmadularu:
It would be great to have this merged (and a new version released) since we are also in the process of upgrading helia.

haydenyoung:
@haadcode I have moved the pinning test into the orbitdb replication. It didn't seem to make sense having a separate test suite for one test.

haadcode (approved):
LGTM 👍

note that this fixes #1163

PR #1200, opened 2024-10-29
fix: Use Helia defaults for gossipsub and circuit relay. Update circu…

haydenyoung:
…it relay with up-to-date configuration.


haadcode (approved):
Lgtm

PR #1221, opened 2025-02-16
Add listener to events.on("error")

julienmalard:
Node.js, by default, throws an error if an EventEmitter emits an "error" event and no other listener is configured. This causes OrbitDB to crash the main thread on Node.js whenever a libp2p error occurs during synchronisation.

### DIFF
--- a/src/sync.js
+++ b/src/sync.js
@@ -133,6 +133,7 @@ const Sync = async ({ ipfs, log, events, onSynced, start, timeout }) => {
    * @instance
    */
   events = events || new EventEmitter()
+  events.on("error", console.log)
 
   timeout = timeout || DefaultTimeout
 


haydenyoung:
Adding for reference; this node issue is documented [here](https://nodejs.org/api/events.html#error-events).

silkroadnomad:
I don't really understand the fix. The question is, do I only need to add error event listener somewhere, or will that be added directly in OrbitDB, in my app, or I am getting something completely wrong.

silkroadnomad:
This is very good news Julian! 

Thanks so much for figuring that out. This issue led to db crashes, in my opinion, which made it inaccessible and causes a lot of headaches.

I am wondering if an interrupted internet connection during an ongoing longer replication / bigger db sync now just leads to an incomplete sync or a sync that is rolled back again or still in a corrupted level db. Also. if I click a button to load a remote db,  it would sync and if I click quockly another view times to load another db, it would probably not finish the replication which then leads to "some sort of stream ended pushable errors" (but now without the app crash)

In case an error occurs, however, it is necessary to close the db so nothing bad happens. 

I am still tapping a bit in the dark here. But it's good we are having all the same challenges.

PR #1222, opened 2025-02-16
Add test for error listener on Sync

julienmalard:


julienmalard:
This test is meant to fail here; it will pass once #1221 is merged.
Unhandled rejections pass in the Mocha tests but crash Node.js scripts (https://nodejs.org/api/cli.html#--unhandled-rejectionsmode).

julienmalard:
Note: This test is incorporated into #1221, and this PR is just provided as reference and should not be merged.

PR #1223, opened 2025-02-21
Remove disconnected peers

julienmalard:


haadcode:
Thanks @julienmalard! This is a good fix 👍

Fid the listing error so the test run and this can be merged.

haadcode (approved):
Lgtm 👍

PR #1224, opened 2025-02-28
Set `runOnLimitedConnection: true` for sync protocol

julienmalard:


aphelionz (dismissed):

haydenyoung:
Looks like test failed with a open leveldb error. Re-ran the test manually and it is succeeding. Do we want to merge or do we want to investigate this leveldb issue further. Looks like it maybe be an inconsistent issue with Level.

julienmalard:
@haydenyoung Any chance this could be merged?

julienmalard:
@abuvanth This will allow OrbitDB to run over circuit relays (essential for most browser-to-browser communication).

PR #1225, opened 2025-03-07
refactor: Pin verified entry.

haydenyoung:


silkroadnomad:
How is the current behaviour regarding to pinning onto the underlying Helia IPFS node when adding data to the log? 
Helia is or was always pinning the data by default? Or is or was the garbage collector removing the pinned OrbitDB data at some point?

haydenyoung:
> How is the current behaviour regarding to pinning onto the underlying Helia IPFS node when adding data to the log? Helia is or was always pinning the data by default? Or is or was the garbage collector removing the pinned OrbitDB data at some point?

Data needs to be explicitly pinned by calling the pin.add function.

When a record is fetched by OrbitDB (during replication), it fetches the data via IPFS's get function but it is not pinned until the record is verified (I.e. it is good data).

PR #1241, opened 2025-11-04
Fix sync error when libp2p nodes are disconnected between reception of head and persistence of entry

julienmalard:
This pull request fixes an error caused by libp2p node disconnection between the reception of a database entry and the persistence of the same entry in the IPFS blockstore. 

This is caused by the following chain of events:

1. The database receives the entry and calls `joinEntry`on the log.
2. The log calls `oplogStore.addVerified`.
3. `addVerified` then calls `await _entries.persist(hash)` on the IPFS blockstore. If, by now, the underlying libp2p nodes are disconnected, the call to `persist` on the IPFS blockstore will hang, even if the OrbitDB node has already received the full entry, because this entry has not yet been added to the local IPFS node and the call to `pin` will attempt to retrieve the entry by its hash from the wider IPFS network, where it is no longer available.

This pull request fixes the issue with two changes:

1. In `log.js`, `oplogStore.addHead` is now called before `oplogStore.addVerified`
2. In `oplog-store.js`, `addHead` now explicitly adds the new head to the `_entries` storage.

### DIFF
--- a/test/database-replication.test.js
+++ b/test/database-replication.test.js
@@ -202,7 +202,6 @@ describe('Database - Replication', function () {
         const peerId = ipfs2.libp2p.peerId
         try {
           const entry = await Entry.decode(message.detail.data)
-          console.log(entry)
           if (entry.hash === expectedEntryHash) {
             entryReceived = true
             await ipfs1.libp2p.hangUp(peerId)


NiKrause:
Is there a way to test such a behaviour without the fix?

